PR #11, opened 2022-02-02
Add microvm kernel config to be a package

GTrunSec:
So, We just need to push the microvm-kernel to our cache server that is enough to speed the building process up.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -78,6 +78,7 @@
             } // args);
           in
             {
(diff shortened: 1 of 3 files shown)


astro:
Good idea!

However, this PR does a lot of whitespace changes which is going to cause hell for me once I'm done with my current refactoring.

GTrunSec:
I got it. I can revert the formatting changes.

astro:
Thank you

Issue #16, opened 2022-02-05
Do not limit interface type to a specific few for all Hypervisors.


GTrunSec:
https://github.com/astro/microvm.nix/blob/main/nixos-modules/microvm/options.nix#L86
The type should be str that more sense. if you would like to assert the type of hypervisor for an interface.  Why not use assertion for different hypervisors.

astro:
Assertions are not very declarative.

I'd like to put all interface types that are available in that enum. For example `"user"` only makes sense for qemu. Other hypervisors should probably refuse that with an explanatory exception.

Thoughts?

GTrunSec:
Sounds good,  but for that in my use [case](https://github.com/GTrunSec/lambda-microvm-hunting-lab/blob/main/hosts/qemu-bridge/microvm.nix#L14). How to set the custom type for qemu, such as bridge?

astro:
In theory, we should add it to the enum. Still, to me it looks like qemu's `bridge` is the same as `tap` in qemu.

In this case I think we should add an interface option for automatic attachment to a bridge, and adapt/reject it for all of the supported hypervisors.

Note that in `examples/microvms-host.nix` I let systemd-networkd take care of attaching the tap interfaces to a virbr0.

astro:
@Neon-44 Unfortunately, the link in the opening post is rather outdated. Could you please open a ticket with an updated description of your issue?

Neon-44:
I am very very sorry. I commented on the wrong Issue. I am an Idiot.

I wanted to comment on this Issue: https://github.com/astro/microvm.nix/issues/260

Should I open a new Issue or just move the Discussion to there?

astro:
That issue is closed, too. Please open a new one.

PR #17, opened 2022-02-05
nixos-modules/microvm/options: add bridge to interface type

GTrunSec:
responses to #16

### DIFF
--- a/nixos-modules/microvm/hypervisor/cloud-hypervisor.nix
+++ b/nixos-modules/microvm/hypervisor/cloud-hypervisor.nix
@@ -39,7 +39,7 @@ in {
          else []) ++
         (if interfaces != []
          then [ "--net" ] ++
-              (map ({ type ? "tap", id, mac }:
+              (map ({ type ? "tap", id, mac, ... }:
                 if type == "tap"
                 then "tap=${id},mac=${mac}"
                 else throw "Unsupported interface type ${type} for Cloud-Hypervisor"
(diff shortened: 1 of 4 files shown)


astro:
Could you please write bridge support or nicely readable exceptions for the other four hypervisors? Being able to switch between them is a big feature of this project.

GTrunSec:
No problem, I will add it by tomorrow.

GTrunSec:
How to run the nested bridge in microvm

```
# /var/lib/microvms/bridge_qemu_2/current/bin/microvm-run       
failed to add interface `tap0' to bridge `enp0s6': Operation not supported
qemu-system-x86_64: -netdev bridge,br=enp0s6,id=vm-qemu2: bridge helper failed
``

**Comment by astro** [comment]
Add the interface with `tunctl`

Does this functionality require running the hypervisor+microvm as root? I dislike putting in such a requirement that will encourage people to run more stuff as root.

The `microvm-tap-interfaces@` systemd unit in the **host** module takes care of creating tap interfaces. I use *systemd-networkd* to attach them to a bridge. I guess there are more solutions to this but attaching interfaces to a bridge can**not** be done by non-root users, right? If so, I think it should not be done by the hypervisor.

**Comment by GTrunSec** [comment]
>  I guess there are more solutions to this, but attaching interfaces to a bridge cannot be done by non-root users, right? If so, I think it should not be done by the hypervisor.

Yes, it requires the root user for a bridge mode. But, I think I can be solved by security.wrapper to `qemu-bridge-helper`(ref to services.libvirtd.qemu.runAsRoot = false;)

the flags will be:
```
-netdev bridge,br=virbr0,helper=/run/wrappers/bin/qemu-bridge-helper,id=vm-qemutap:
```
I'm not sure that I did a correct flag of the helper.

furthermore, How do you use ssh to connect tap microm in different networks?
port forward to hostfwd=tcp::2222-:22 ? I'm sorry about the weak knowledge of VM I am.

```
ssh -p  2222 localhost ?
```

PR #39, opened 2022-05-04
doc: fix simple network example

Sohalt:


astro:
oopsie, thank you very much!

PR #40, opened 2022-05-04
nixos-modules/host: enable Kernel Same-Page Merging

astro:
KSM seems like a sensible feature to save memory when running many MicroVMs. Yet I get zero shared pages with both QEMU and Cloud-Hypervisor.

### DIFF
--- a/nixos-modules/host.nix
+++ b/nixos-modules/host.nix
@@ -198,5 +198,8 @@ in
     environment.etc."qemu/bridge.conf".text = lib.mkDefault ''
       allow all
     '';
+
+    # Enable Kernel Same-Page Merging
+    hardware.ksm.enable = true;
   };
 }


jkachmar:
haven't tested this yet, but it looks like `hardware.ksm.enable = true` doesn't indicate to libvirt that KSM should be enabled (e.g. by writing `AUTO` to `/etc/defaults/qemu-kvm`).

it looks like `cloud-hypervisor` has to be explicitly invoked with `mergeable=on` (per [the docs](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/docs/memory.md#mergeable)) to work with KSM.

i'll try to update here when/if i get a chance to try this out.

astro:
Note that we don't go through libvirt.

jpds:
Also spotted: https://gitlab.com/virtio-fs/virtiofsd/-/issues/109

astro:
Merged with 42ac0e7 and 4caeec0

PR #51, opened 2022-07-06
Warn if trying to use nix store optimisation with writableStoreOverlay

### LINKED ISSUE #50: `nix.optimise.automatic` and `nix.auto-optimise-store` do not work with `writeableStoreOverlay`
`nix.optimise.automatic` and `nix.auto-optimise-store` try to create hardlinks accross the layers of the overlayfs, which fails.

I think we should either have an `assert` with a helpful error message, or `lib.mkForce` these options to `false`, when using an overlay.

Sohalt:
Nix store optimisation tries to create hardlinks between store paths
with identical contents. This fails when trying to create a hardlink
between different layers of the overlay.

Closes #50

### DIFF
--- a/nixos-modules/microvm/system.nix
+++ b/nixos-modules/microvm/system.nix
@@ -5,6 +5,12 @@ let
   }) defaultFsType withDriveLetters;
 in
 {
+  assertions = [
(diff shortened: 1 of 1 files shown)


astro:
Thank you very much!

PR #56, opened 2022-07-11
microvm-install fixes

Sohalt:
When run multiple times, ln -sf ${runner} current tries to create
nested symlinks inside of the directory linked to by the previous
invokation.

Without -h, chown dereferences symlinks and tries to change the owner of
files in the nix store.


astro:
Thank you very much

PR #58, opened 2022-09-03
Avoid nix caching problem when listing microvm's newest version

SuperSandro2000:


astro:
Given that it can take over 10s to evaluate a `lib.nixosSystem` and there can be many of them in one `microvm -l` run, I am looking for a sweet spot between 10s and the default of 3600s.

Perhaps we should make it equal to time that has elapsed since the script was started?

Issue #61, opened 2022-11-03
Make the modules enable-able?


nazarewk:
It currently does not look like you could import `host`/`microvm` nixos modules without triggering side effects.

It would be great to have `enable` flag (could default to true) to hide all changes behind each of those.

PR #62, opened 2022-11-03
make modules opt-out after import

nazarewk:
fixes https://github.com/astro/microvm.nix/issues/61

### DIFF
--- a/nixos-modules/microvm/stage-1.nix
+++ b/nixos-modules/microvm/stage-1.nix
@@ -8,67 +8,71 @@ let
     else "$targetRoot" + (
       # find share with host's /nix/store
       builtins.head (
-        builtins.filter ({ source, ... }:
-          source == "/nix/store"
-        ) config.microvm.shares
+        builtins.filter
+          ({ source, ... }:
+            source == "/nix/store"
+          )
+          config.microvm.shares
       )
     ).mountPoint;
-in {
+in
+{
   config = lib.mkIf config.microvm.guest.enable {
-  system.build.microvmStage1 = pkgs.substituteAll rec {
-    src = ./stage-1-init.sh;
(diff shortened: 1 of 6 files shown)


nazarewk:
this is not side-effect-free, but good enough (only enables imported minimal profile).

nixos-modules/microvm/stage-1.nix
```diff
@@ -69,4 +70,5 @@ in {
     fsType = "tmpfs";
     options = [ "size=50%,mode=0755" ];
   };
+};
```
astro: Lines in between these curly braces need to be indented.

nixos-modules/microvm/system.nix
```diff
@@ -90,4 +92,5 @@ in
       generators = { systemd-gpt-auto-generator = "/dev/null"; };
     };
 
+};
```
astro: Likewise

astro (changes requested):
Thank you. This looks good to me except for the indentation.

May I ask about the use-case for this change? I want to avoid breaking it in the future.

nazarewk:
I thought about formatting it with `nixpkgs-fmt`, but it made MANY changes.

About my use case: I have a [repo holding all my configs in opt-in modules](https://github.com/nazarewk-iac/nix-configs), so i can pick only relevant pieces for different configs/machines. It works fine except for things using "plain" configs not hidden behind (eg: [this importing qemu guest from nixpkgs](https://github.com/nazarewk-iac/nix-configs/blob/b8f5532608650f57f1f14ecd05a0e9b07887d088/machines/hetzner/wg-0/default.nix#L2-L6))

astro:
git's word-diff didn't work well for me regardless of the chosen algorithm. Therefore I indented myself.

Thank you!

PR #71, opened 2023-02-01
lib/runners/crosvm: Clean-up control socket

mikatammi:
crosvm doesn't properly clean the control socket when it stops, ensure there is no control socket, to avoid "Address already in use" when it is trying to start again.

Signed-off-by: Mika Tammi <mika.tammi@unikie.com>

### DIFF
--- a/lib/runners/crosvm.nix
+++ b/lib/runners/crosvm.nix
@@ -10,6 +10,8 @@ let
   mktuntap = pkgs.callPackage ../../pkgs/mktuntap.nix {};
   interfaceFdOffset = 3;
 in {
+  preStart = "rm -f ${socket}";
+
   command =
     if user != null
     then throw "crosvm will not change user"


phaer:
Nice catch, but I am curious: Why `preStart` and not `postStop` or the shutdown command? Wouldn't this still leave the socket around if you just stop the machine?

mikatammi:
> Nice catch, but I am curious: Why `preStart` and not `postStop` or the shutdown command? Wouldn't this still leave the socket around if you just stop the machine?

Yes it would. But if the system crashes, then the `postStop` and shutdown commands will not be run, and the next time system boots, there will be this error again.

I have nothing against adding the same thing to stop-commands though, in addition to making sure it is possible to start the crosvm.

What I was also considering: NixOS has the /run tmpfs-filesystem, which is naturally clean when starting the system. Should we consider that to be the default path for sockets? (It would eliminate the need for this rm-patch altogether)

astro:
Thank you!

> What I was also considering: NixOS has the /run tmpfs-filesystem, which is naturally clean when starting the system. Should we consider that to be the default path for sockets? (It would eliminate the need for this rm-patch altogether)

Did you know that they are configurable? It's just the defaults that are simple paths. I'll consider any improvement that covers both use-cases (starting a MicroVM interactively + through systemd from /var/lib/microvms).

Issue #69, opened 2023-01-27
Kernels possibly taken from wrong Nix store path


mikatammi:
I'm trying to make MicroVM.nix work on aarch64-linux host architecture with aarch64-linux guests

For me it looks like kernel is taken from wrong path: I see `"${kernel.dev}/vmlinux"` all over the place, I think this should be `"${kernel.out}/Image"` instead

That seems to not cause errors on x86_64 host, but on aarch64 I can see clear errors about kernel image being wrong type with crosvm, and QEMU just doesn't do anything

astro:
Sorry for the sad state of aarch64 support here. I got tired of waiting for my RPi4.

I am going to be very happy about PRs.

mikatammi:
There probably will be few :)

mikatammi:
What concerns me about this vmlinux vs. Image -thing, is that is there some benefit for using the vmlinux-image on x86_64? My educated guess is that the vmlinux has the debugging information, but if one does not do kernel debugging, is there any benefit for using vmlinux instead of Image.

For the PR I'm just thinking should I make this conditional based on system architecture, or just change to Image altogether?

astro:
One of the hypervisors wanted the Image, it couldn't load vmlinux.

On a side note I am currently trying to build https://github.com/cloud-hypervisor/rust-hypervisor-firmware in the hope that we can boot ELF kernel + initramfs on all of the five hypervisors.

mikatammi:
You are right, we need ELF for x86_64 and the (b)zImage for aarch64. I guess there are really no other architectures with virtualization right now, maybe PPC and also RISC-V in the future

PR #72, opened 2023-02-01
Make QEMU, CrosVM, Cloud-Hypervisor, Firecracker and kvmtool work on both x86_64 and aarch64

mikatammi:
Created this to open discussion. Relates to #69 

On x86_64, some hypervisors are able to load the ELF-formatted vmlinux binaries, which have debugging symbols included. However, this does not apply currently for aarch64-linux.

Signed-off-by: Mika Tammi <mika.tammi@unikie.com>

### DIFF
--- a/lib/runners/cloud-hypervisor.nix
+++ b/lib/runners/cloud-hypervisor.nix
@@ -5,7 +5,7 @@
 }:
 
 let
-  inherit (pkgs) lib;
+  inherit (pkgs) lib system;
   inherit (microvmConfig) vcpu mem balloonMem user interfaces volumes shares socket devices hugepageMem;
 
   # balloon
@@ -13,6 +13,11 @@ let
 
   # Transform attrs to parameters in form of `key1=value1,key2=value2,[...]`
   opsMapped = ops: lib.concatStringsSep "," (lib.mapAttrsToList (k: v: "${k}=${v}") ops);
+
+  kernelPath = {
+    x86_64-linux = "${kernel.dev}/vmlinux";
+    aarch64-linux = "${kernel.out}/${pkgs.stdenv.hostPlatform.linux-kernel.target}";
+  }.${system};
 
   # Attrs representing CHV mem options
   memOps = opsMapped ({
@@ -71,7 +76,7 @@ in {
         "--watchdog"
         "--console" "tty"
         "--serial" "pty"
-        "--kernel" "${kernel.dev}/vmlinux"
+        "--kernel" "${kernelPath}"
         "--cmdline" "console=hvc0 console=ttyS0 reboot=t panic=-1 ${toString microvmConfig.kernelParams}"
         "--seccomp" "true"
         "--memory" memOps
--- a/lib/runners/crosvm.nix
+++ b/lib/runners/crosvm.nix
@@ -5,11 +5,15 @@
 }:
 
 let
-  inherit (pkgs) lib;
+  inherit (pkgs) lib system;
   inherit (microvmConfig) vcpu mem balloonMem user interfaces volumes shares socket devices;
   inherit (microvmConfig.crosvm) pivotRoot extraArgs;
   mktuntap = pkgs.callPackage ../../pkgs/mktuntap.nix {};
   interfaceFdOffset = 3;
+  kernelPath = {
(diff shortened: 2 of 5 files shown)


mikatammi:
I tried to Google the stuff (why to use vmlinux), and the only practical reason to use the vmlinux (instead of standard compressed image) I found was the debugging symbols.

I guess also if you start 10 000 microvms, the few CPU cycles used for kernel decompression might start to count, but for the "normal usecase" my guess is the difference is marginal. I saw some benchmarks from 2017 and the gzip decompression took ~50ms. (And the kernel has other config-options to optimize for Virtual Machine usage and this is already a whole different discussion.)

Dan-Theriault:
I tried using this on x86_64 crosvm and I get an error:

```
crosvm[1975471]:  [src/main.rs:98] crosvm has exited with error: failed to open kernel image /nix/store/lwv517nfbn5jnkylrrl9g9nz79r5aind-linux-6.1.8/Image: No such file or directory (os error 2)
```

There is `bzImage` but not `Image` file within that store path.

astro:
Please give me some more time for testing. I wish I had a stable CI setup for more than the main branch. Unfortunately, Github actions don't expose nested KVM. Suggestions welcome.

mikatammi:
Converting this back to draft. This change seems to break x86_64-linux because there the kernel is called bzImage and not Image

mikatammi:
I wonder what would be a good constinent way to get the kernel filename. Should we test for all possibilities?

astro:
Yeah, this has to go behind a check for x86_64/aarch64.

mikatammi:
By investigating nixpkgs further, I found the place where the actual image name is defined:
https://github.com/NixOS/nixpkgs/blob/nixos-22.11/pkgs/os-specific/linux/kernel/generic.nix#L134-L135

```
nix-repl> legacyPackages.aarch64-linux.hostPlatform.linux-kernel.target
"Image"
```

```
nix-repl> legacyPackages.x86_64-linux.hostPlatform.linux-kernel.target
"bzImage"
```

(Also found out it's common misconception, b in bzImage does not mean bzip2 compression, but "big")

So we could probably just use the target variable like `${kernel.out}/${kernel.target}`

mikatammi:
Tested with crosvm and works for both on x86_64-linux and aarch64-linux

mikatammi:
The kernelTarget wasn't actually exposed so I had to use the long `${pkgs.stdenv.hostPlatform.linux-kernel.target}`

astro:
Please rebase :)

Are you certain that currently the same `pkgs.stdenv.hostPlatform.linux-kernel.target` will work for all 5 hypervisors, also on x86_64-linux? That was not my experience.

mikatammi:
I can try to do some checks, I have mainly tested things on aarch64-linux

mikatammi:
I've been fixing things and this starts to become quite large. I'm splitting this to series of smaller Pull Requests to fix simple things first. QEMU and Firecracker need more working. With QEMU it just about the parameters but Firecracker is more problematic. First of all, nixpkgs has wrong hash for the Firecracker's package. Then after fixing the hash, it doesn't start on aarch64-linux. It does work fine on x86_64-linux though

mikatammi:
First of them here: #85

mikatammi:
There's now:
* #85 for crosvm and cloud-hypervisor fixes
* #86 for kvmtool
* #87 for firecracker
* QEMU will be done later

mikatammi:
Actually all the smaller Pull Requests are now part of this Pull Request

mikatammi:
@astro Ready for review now!

mikatammi:
I tested that all QEMU, CrosVM, Cloud-Hypervisor and Firecracker work on both x86_64 and aarch64. kvmtool package simply isn't supported on aarch64-linux in nixpkgs and even Firecracker's package in nixos-22.11 has wrong hash which I had to change

mikatammi:
I also created a PR to update the kvmtool package in nixpkgs: https://github.com/NixOS/nixpkgs/pull/225450

This makes it work on both aarch64-linux and x86_64-linux!

astro:
That's pretty awesome. Thank you for your efforts!

PR #75, opened 2023-03-07
CHV: fix vfio perm and limits + hugepages

snpschaaf:
Changes regarding cloud-hypervisor (CHV):
  - change the permission of the vfio dev  
  - set memlock in limits
  - add hugepage option 

@astro if you wish i could split this PR into separate ones.


astro:
Looks very well to me. Thank you.

astro:
I think you may have mistakenly written `mem` instead of `size` in the `--memory` argument. I pushed a fix.

snpschaaf:
> I think you may have mistakenly written mem instead of size in the --memory argument. I pushed a fix.

You'r right - bad mistake :see_no_evil: Thx for fixing it.

PR #85, opened 2023-04-08
Fix crosvm and cloud-hypervisor kernel paths

mikatammi:
There are now different kernel paths for x86_64-linux and aarch64-linux


mikatammi:
Everything is in #72

PR #86, opened 2023-04-08
kvmtool: Change kernel path generation method

mikatammi:
Change kvmtool to use the more portable
${pkgs.stdenv.hostPlatform.linux-kernel.target} as a kernel image filename, instead of the hardcoded bzImage.


mikatammi:
Everything is in #72

PR #90, opened 2023-04-14
qemu: Require PCI for PCI passthrough on x86_64

mikatammi:
vfio-pci requires PCI bus for qemu

Otherwise you get errors like:
`qemu-system-x86_64: -device vfio-pci,host=0000:00:14.3,multifunction=on: No 'PCI' bus found for device 'vfio-pci'`


astro:
Thank you very much!

Issue #80, opened 2023-03-29
Support MACVTAP as an alternative to classical TUN/TAP


oddlama:
I've been using MACVTAP networking instead of classical TUN/TAP networking on my VMs for some time and would like to do the same when finally migrating to microvm.nix. MACVTAPs have some advantages over TUN/TAP interfaces that I learned to appreciate, but also some downsides so this probably isn't a replacement for TUN/TAP networking.

Put shortly, it is a much simpler driver that eliminates a whole layer by giving each guest it's own physical address. This _vastly_ reduces the processing needed on the host and allows for high-throughput between host and guests without putting additional stress on the CPU. It can also be used to fully isolate host and guest traffic (by operating in private or passthrough mode and not creating a vtap for the host), if so desired. While IBM does report [improvements in performance between 10-50%](https://www.ibm.com/docs/en/linux-on-systems?topic=cons-macvtap-driver-considerations), I can personally only attest to the reduction in required compute. Thus I believe it would make for a great feature for microvm.nix.

Also, see [the diagrams in this redhat guide](https://developers.redhat.com/blog/2018/10/22/introduction-to-linux-interfaces-for-virtual-networking#macvtap) for a visual comparison between MACVTAP and TAP/TUN architectures. As a quick reference, here's what a typical setup using systemd-networkd would entail:

Create a macvtap device in bridge mode (to allow guest-guest networking):
```desktop
# /etc/systemd/network/br-eth0.netdev
[NetDev]
Name=br-eth0
Kind=macvtap

[MACVTAP]
Mode=bridge
```

The host's networking is configured on the macvtap interface to allow host-guest networking:
```desktop
# /etc/systemd/network/br-eth0.network
[Match]
Name=br-eth0

[Network]
# host network configuration, whatever it would otherwise want for eth0 directly
Address=192.168.1.1/24

# This is necessary to prevent an unconfigured bridge when the attached physical
# device fails to come up. If for example the LAN cable is removed from the real
# interface, we still want a static ip on the bridge to allow our virtual machines
# to function correctly.
ConfigureWithoutCarrier=yes
```

Attach the macvtap to eth0
```desktop
# /etc/systemd/network/eth0.network
[Match]
Name=eth0

[Network]
MACVTAP=br-eth0
```

Each guest can now get it's own macvtap interface. Probably depends on the hypervisor on how to set that up - I previously used libvirt which does that on demand.

---

Additional relevant resources:

[...]

PR #93, opened 2023-04-19
Create macvtap interface link

megheaiulian:
Auto create macvtap interfaces with a service similar to tap interfaces. 

- Adds a `link` option to interfaces specifying which interface the auto-created macvtap should link to.
- Adds service that will create macvtap interfaces on the host

Example configuration:
```
microvm.interfaces = [{
   type="macvtap";
   id="macvtap1";  # name of the newly created interface on the host
   link="eno2";  # name of an existing interfaces that the macvtap operates on
   mac="00:00:00:00;00"; # the mac address of the macvtap interface
}];
```

Related to #80

### DIFF
--- a/nixos-modules/microvm/options.nix
+++ b/nixos-modules/microvm/options.nix
@@ -194,6 +194,12 @@ in
               Interface name on the host
             '';
           };
+          link = mkOption {
+            type = str;
+            description = ''
+              Attach network interface to host interface for type = "macvlan"
+            '';
+          };
           bridge = mkOption {
             type = nullOr str;
             default = null;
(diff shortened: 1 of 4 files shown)


nixos-modules/microvm/options.nix
```diff
@@ -194,6 +194,12 @@ in
               Interface name on the host
             '';
           };
+          link = mkOption {
```
astro:
```suggestion
          macvtap.link = mkOption {
```

astro (changes requested):
Awesome! This is even better than if I would have done it myself. :-)

In my experience, the macvtap mode (`private` | `vepa` | `bridge` | `passthru`) is important. Could you please add that? I think it would make sense to move the new options to `microvm.interfaces.*.macvtap.{link,mode}`.

PR #95, opened 2023-04-21
fix(macvtap): correctly setup mode

megheaiulian:
Correctly pass `mode $mode` as arguments array instead of a string.
Ref #93 #80


astro:
Thanks for taking care!

nephalemsec:
Apologies for reviving a dead thread. But if work is/was being done on macvtap does this mean macvlan could potentially be on the cards?

astro:
@nephalemsec macvlan is in the cards. :spades: :hearts: :clubs: :diamonds:

Issue #94, opened 2023-04-21
option to autoinstall/autoupdate declarative vms


megheaiulian:
Currently microvm does not install updates to a vm if the vm already exists in state dir.
To push/deploy the new configuration user would need to delete the old vm or `microvm  -u`.
Since by default the vms is using tmpfs and all its state is ephemeral I think it would make sense to add a `autoupdate` or `autoinstall` option that does the install and overwrites the vm on change.
This is prevented by `ConditionPathExists` https://github.com/astro/microvm.nix/blob/d578ed45d3528c42ab4b12edf6d1512a7bd35714/nixos-modules/host.nix#L117 and the new boolean option would just disable that.

megheaiulian:
@astro Let me know if you think something like that makes sense? Would work nicely for stateless vms.

astro:
Thank you for giving this some thought. The downside of this approach is that a host reboot will roll back a VM that was previously updated with `microvm -u`. Then again, that probably won't be used if the autoupdate option is enabled?

megheaiulian:
When using the declarative approach it is not very pleasant to have to run `microvm -u`.
Ideally i'd like to do a change in the configuration, rebuild and get the latest configuration up and running.

astro:
I recently added the `autorestart` option, and I am really unhappy about its naming because it does not convey much of its purpose. People may use it now, so renaming it would be bad. Do you have any ideas for a better name than `autoupdate`? To me, *autoupdate* sounds like true automated updates.

Actually, the `"install-microvm-${name}"` service can already access a MicroVM's config *because* in this case it is evaluated together with the host. Move the `let` block under `script` up a level, then this new option can become individually configurable per MicroVM.

I would like to add that instead of the option we could get some nice automatic behavior here if we were only able to tell which MicroVM config is newer. Everything under the store is 1970-01-01 but maybe we could export the `flake.sourceInfo.lastModified` into the runner package? Would that work for you?

astro:
Thank you for elaborating on the topic.

What this really would do is kind of `alwaysInstall` because it causes the install-microvm service to be run on every boot and `nixos-rebuild switch`. Right?

megheaiulian:
Tried to implement this but not sure how we could both mark that the install-microvm service is part of microvm-%i (so that it start when microvm-%i starts) but also when it restarts make microvm-%i restart too.

astro:
[...]

Issue #92, opened 2023-04-17
add the ability to supply a config instead of just a nixosConfigurations flake


bolives-hax:
Hey as of the way my nixos configurations are structured. It is not possible for me to supply the system configuration for the vm over the route below.

```nix
{
vms.${name}.flake = self;
}
```
as of the following line of code it seems https://github.com/astro/microvm.nix/blob/d578ed45d3528c42ab4b12edf6d1512a7bd35714/nixos-modules/host.nix#L122

Why could this matter?.  Let me suggest a use case: 

Lets say I want to supply certain information to my "vm_httpd" vm definition. Such as a domain names it makes use of when generating http. Instead of declaring  that in ``/etc/nixos#nixosConfigurations.vm_httpd`` it would seem much more cool to generate it from within my primary config ``/etc/nixos#nixosConfigurations.bigServer`` via a convenience function called 
```nix
mkHttpdConfig = {domainName}: {
   config."...".domain = domainName;
} 
```
or something like that. This starts being useful as soon as one wishes to spawn multiple  vm with nearly identical configuration values.  Below can be seen how id roughly imagine that and how I already do it for my containers at least:

```nix
{
inputs.myWebsite.url = "github:user/repo";
# ...
nixosConfigurations.bigServer = nixosSystem {
   system="x86_64-linux";
   modules = let
        webDomain = "www.domain.de";
        templates = import ./templates; 
   in [
       microvm.host
       ({...}: {
          micro.vms.httpd_a.config = (templates.vm_httd_public {
            inherit domain;
            serveData = myWebsite;
          });
          micro.vms.httpd_a.config = (templates.vm_httd_tor {
             serveData = myWebsite;
             torConfig = secrets.torCreds01;
          });
       })
   ];
}
```

Being able to express large parts of my nixos configurations as functions that then end up producing valid configurations to actually be build would allow for much more dynamic configuration approaches. Less duplication and cleaner code. 

I personally don't seem to see any issues with implementing this alternative behavior, though maybe im missing something or it goes against the design decisions taken by the author so I'd rather be interested in the current take on my suggestion.  If there is any interest in this feature and the option of a potential change being merged exist I would also make the change myself.

oddlama:
@bolives-hax I just had the same problem and implemented this quickly. If you are open to trying the liked PR, I'd appreciate your feedback and whether it works as you expect it to.

astro:
[...]

PR #96, opened 2023-05-08
Implement fully declarative approach via direct `config` as an alternative to specifying `flake`

oddlama:
This PR adds an alternative option `vms.<name>.config` that can be used instead of specifying `vms.<name>.flake`. This way of doing configuration doesn't impose any restrictions on the user's flake, but can only be used declaratively. This is similar to how e.g. nixos-container is doing this.

- This addresses #92 
- and also adds `declarativeUpdates` for #94 that restarts VMs when the toplevel is changed and the host is `nixos-rebuild --switch`-ed

This PR also renames an option which might or might not be desirable. I'm not overly familiar with the codebase yet, so I "hacked" together something to make the microvm list command work again, since it depended on the `flake` file to determine the vm's toplevel. This should probably be revised, feedback is always welcome.

Example usage:

```nix
{ microvm, pkgs, ... }: {
  microvm.declarativeUpdates = true;
  microvm.restartIfChanged = true;
  microvm.vms.test.config = {
    imports = [microvm.nixosModules.microvm];
    # typical config ...
  };
}
```

This also allows merging multiple definitions, so specifying `config` again in another module will just work:

```nix
{
  microvm.vms.test.config = {
    services.nginx.enable = true;
  };
}
```

You can also specify another `pkgs` set and `specialArgs`, in case the guest should use a different system, or needs certain special arguments:

```nix
{ microvm, pkgs, ... }: {
  microvm.vms.test = {
    inherit pkgs; # optional (default to host's pkgs), used to selects the guest system and nixpkgs package set.
    specialArgs = {}; # optional, passed to nixosSystem
    config = {
      imports = [microvm.nixosModules.microvm];
      # typical config ...
    };
  };
}
```

Remarks:

All outstanding issues have been addressed. See comments below.

- ~~currently an assertions ensures `delcarativeUpdates -> restartIfChanged`. This could maybe be merged into one option, or left out by implying declarativeUpdates when using `vms.<name>.config`.~~
- ~~I'm not familiar enough with all the systemd services that are created, so I'm not 100% confident that `restartIfChanged` is always doing what it should be doing~~
- ~~declarative vm's are not yet restarted when their toplevel is changed. Waiting on #94 to see how we will solve that.~~

### DIFF
--- a/nixos-modules/host.nix
+++ b/nixos-modules/host.nix
@@ -50,6 +50,14 @@ in
       '';
     };
 
+    declarativeUpdates = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Deploys updates to existing microvms automatically when the host is rebuilt.
+      '';
+    };
+
     autostart = mkOption {
       type = with types; listOf str;
       default = [];
@@ -60,11 +68,11 @@ in
       '';
     };
 
-    autorestart = mkOption {
+    restartOnChange = mkOption {
       type = types.bool;
       default = false;
       description = ''
-        Restart MicroVM services on `nixos-rebuild switch` of the host?
+        Restart MicroVM services if the systemd services are changed.
       '';
     };
   };
@@ -114,7 +122,7 @@ in
         partOf = [ "microvm@${name}.service" ];
         wantedBy = [ "microvms.target" ];
         # only run if /var/lib/microvms/$name does not exist yet
-        unitConfig.ConditionPathExists = "!${stateDir}/${name}";
+        unitConfig.ConditionPathExists = lib.mkIf (!config.microvm.declarativeUpdates) "!${stateDir}/${name}";
         serviceConfig.Type = "oneshot";
         script =
           let
--- a/nixos-modules/host.nix
+++ b/nixos-modules/host.nix
@@ -18,21 +18,85 @@ in
     };
 
     vms = mkOption {
-      type = with types; attrsOf (submodule ({ ... }: {
+      type = with types; attrsOf (submodule ({ config, name, ... }: {
         options = {
+          config = mkOption {
+            description = lib.mdDoc ''
+              A specification of the desired configuration of this
+              MicroVM, as a NixOS module.
+            '';
+            default = null;
+            type = nullOr (lib.mkOptionType {
+              name = "Toplevel NixOS config";
+              merge = loc: defs: (import "${toString config.pkgs.path}/nixos/lib/eval-config.nix" {
+                modules =
+                  let
+                    extraConfig = {
+                      _file = "module at ${__curPos.file}:${toString __curPos.line}";
+                      config = {
+                        networking.hostName = lib.mkDefault name;
+                      };
+                    };
+                  in [
+                    extraConfig
+                    # TODO missing self:
+                    # (import ./microvm self)
+                  ] ++ (map (x: x.value) defs);
+                prefix = [ "microvm" name ];
+                inherit (config) specialArgs pkgs;
+                inherit (config.pkgs) system;
+              }).config;
+            });
+          };
+
+          pkgs = mkOption {
+            type = types.unspecified;
+            default = pkgs;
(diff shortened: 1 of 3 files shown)


astro:
Cool, thanks. I guess this would make this project usable to even more contributors. :tada: 

Did you consider extending the non-flake story to imperative microvm management with the `microvm` command, and maybe even packages? No need to bloat this PR just yet, I going to review as is...

nixos-modules/host.nix
```diff
@@ -112,26 +176,30 @@ in
...
           in
           ''
             mkdir -p ${stateDir}/${name}
             cd ${stateDir}/${name}
 
             ln -sTf ${runner} current
-
+            chown -h ${user}:${group} . current
+          '' + lib.optionalString (flake == null) ''
+            echo ${microvmConfig.system.build.toplevel} > toplevel
```
astro: Make a link: `ln -s`

pkgs/microvm-command.nix
```diff
@@ -166,8 +166,12 @@ writeScriptBin "microvm" ''
           CURRENT_SYSTEM=$(readlink "$DIR/current/share/microvm/system")
           CURRENT=''${CURRENT_SYSTEM#*-}
 
-          FLAKE=$(cat "$DIR/flake")
-          NEW_SYSTEM=$(nix --option narinfo-cache-negative-ttl 10 eval --raw "$FLAKE#nixosConfigurations.$NAME.config.system.build.toplevel")
+          if [ -e "$DIR/toplevel" ]; then
+            NEW_SYSTEM=$(cat "$DIR/toplevel")
```
astro:
```suggestion
            NEW_SYSTEM=$(readlink "$DIR/toplevel")
```

astro: Now that we add support to the `microvm` command, should `$DIR/toplevel` get deleted when user updates a MicroVM?

oddlama:
> Did you consider extending the non-flake story to imperative microvm management with the `microvm` command, and maybe even packages? No need to bloat this PR just yet, I going to review as is...

Sure, but I wasn't sure what the preferred implementation would be. Since I'm not overly familiar with the internals of microvm yet I refrained from implementing more than necessary for now. But I'm open to extending this. In either case I'd like to address the remarks before considering this final. Is there a matrix room for microvm yet?

astro (changes requested):

pkgs/microvm-command.nix
oddlama: See updated comment below

oddlama:
I've now restructured this a bit to resolve some of the outstanding issues:

- The `declarativeUpdates` option was removed. Fully-declarative microvms are now distinguished from declaratively deployed but imperatively manged ones simply by checking whether they used `config` or `flake` in the definition.
- Moved the restartIfChanged option to a per-vm configuration. Purely imperative microvms couldn't be affected by this anyway so this is more granular and more correct.
- The `restartIfChanged` directive has been removed from many units, since `partOf` already accomplishes the same. Instead the master group unit `microvm@<name>` is now restarted if necessary.
- The microvm command will refuse to update fully-declarative vms.
- Added docs :D

This should address all your comments and my original remarks. Handling of the master unit restart is still accomplished by a hacky practice IMO, but I cannot think of another way to do it - suggestions are welcome.

oddlama:
I'd say this is ready now, but for the automatic restarts to take effect we will have to wait for https://github.com/NixOS/nixpkgs/pull/232230.

This PR adds deterministic restart on switch behavior for declarative MicroVMs, but not for imperative MicroVMs. If we want users to be able to select whether their imperative microvms should be restarted if the base service changes, this will have to be added as a separate per-instance template unit override. Probably the `microvm` command should create that drop-in, but this is out of scope for this PR.

astro:
I rebased this myself and added a few commits. Thank you!

oddlama:
Damn, now the GPG signatures are lost 😛 
But thanks for working on this ❤️

PR #104, opened 2023-06-06
Don't restart all microvms when deploying host

SuperSandro2000:


oddlama:
This change broke restarting of fully declarative microvms.

`restartIfChanged` is implemented as an opt-out which adds `X-RestartIfChanged=false` to the service definition iff `restartIsChanged = false`. This means that because you set this to `false` on the template service definition, it broke this override statement.

https://github.com/astro/microvm.nix/blob/484e6e2209a0ead8ea43a9a79b193026026becfc/nixos-modules/host.nix#L217-L219

(Setting it to true here has no effect since this doesn't do anything if it is true)
One option to fix this might be adding `X-RestartIfChanged=true` to the per instance override manually.

SuperSandro2000:
please see #110

Issue #106, opened 2023-06-12
 Help configuring macvtap interface


illode:
I've been trying to get macvtap working for a bit now, but the generated interface/ip always points to the host.
The vm config:
```nix
let 
  name = "test-vm";
in { pkgs, microvm, ... }: {
  imports = [ microvm.host ];
  
  microvm.vms.${name} = {
    config = {
      #imports = [ microvm.microvm ];
      networking.hostName = name;
      networking.firewall.enable = false;
      services.openssh = {
        enable = true;
        settings.PermitEmptyPasswords = "yes";
        settings.PermitRootLogin = "yes";
      };
      users.users.root.password = "";
  
      microvm.hypervisor = "cloud-hypervisor";
      microvm.shares = [{
        source = "/nix/store";
        mountPoint = "/nix/.ro-store";
        tag = "ro-store";
        proto = "virtiofs";
      }];
      microvm.interfaces = [{
        type = "macvtap";
        id = "${name}-macvtap";
        mac = "02:00:00:00:00:01";
        macvtap = {
          link = "ConnectX3";
          mode = "bridge";
        };
      }];
      system.stateVersion = "22.11";
    };
  };
}
```
The ConnectX3 link is defined in a separate file:
```nix
{
  systemd.network.enable = true;
  
  systemd.network.links."10-ConnectX3" = {
    matchConfig.MACAddress = "ec:0d:9a:19:d9:a0";
    linkConfig.Name = "ConnectX3";
  };
}
```
The `microvm` arg is passed from `flake.nix` with `specialArgs.microvm = microvm.nixosModules;`

`ip link` on the host:
```
❯ ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: ConnectX3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP mode DEFAULT group default qlen 1000
    link/ether ec:0d:9a:19:d9:a0 brd ff:ff:ff:ff:ff:ff
41: test-vm-macvtap@ConnectX3: <BROADCAST,MULTICAST,ALLMULTI,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 500
[...]

PR #107, opened 2023-06-14
Remove un-needed i8042 device from qemu command line

jkuro-tii:
The i8042 device is added to qemu command line for all platforms, whereas it's needed only for x86. It's already added for the platform in line #106.


mikatammi:
Looks good, this will fix the QEMU on aarch64-linux again

PR #110, opened 2023-06-18
Fix microvm.vms.$NAME.restartIfChanged

SuperSandro2000:


nixos-modules/host.nix
```diff
@@ -224,6 +224,9 @@ in
         restartTriggers = [guestConfig.system.build.toplevel];
         overrideStrategy = "asDropin";
       };
+      "microvm-tap-interfaces@${name}" = microvmConfig.restartIfChanged;
+      "microvm-pci-devices@${name}" = microvmConfig.restartIfChanged;
+      "microvm-virtiofsd@${name}" = microvmConfig.restartIfChanged;
```
oddlama: What are you trying to accomplish? Is assigning a boolean to a service definition even valid?
Since these are instance overrides you also need to specify an override strategy like I did above.

I think to fix the root issue you would need to explicitly add `X-RestartIfChanged` to the service or am I missing something?

SuperSandro2000: I should also write the inherit and good point with the dropin.

oddlama: There still is the issue that what you are doing currently won't work if `restartIfChanged = false` on the template parent (which your previous PR introduced).
See https://github.com/NixOS/nixpkgs/blob/c1bca7fe84c646cfd4ebf3482c0e6317a0b13f22/nixos/lib/systemd-lib.nix#L378-L380
it only sets `X-RestartIsChanged=false` and never `X-RestartIsChanged=true` which causes the actual bug here.

nixos-modules/host.nix
```diff
@@ -224,6 +224,18 @@ in
...
+      "microvm-tap-interfaces@${name}" = {
+        inherit (microvmConfig) restartIfChanged;
+        overrideStrategy = "asDropin";
+      };
+      "microvm-pci-devices@${name}" = {
+        inherit (microvmConfig) restartIfChanged;
+        overrideStrategy = "asDropin";
+      };
+      "microvm-virtiofsd@${name}" = {
+        inherit (microvmConfig) restartIfChanged;
```
SuperSandro2000:
```suggestion
        restartIfChanged = [ "" inherit (microvmConfig) restartIfChanged];
        overrideStrategy = "asDropin";
      };
      "microvm-pci-devices@${name}" = {
        restartIfChanged = [ "" inherit (microvmConfig) restartIfChanged];
        overrideStrategy = "asDropin";
      };
      "microvm-virtiofsd@${name}" = {
        restartIfChanged = [ "" inherit (microvmConfig) restartIfChanged];
```
I don't really have something to test this without causing downtimes for others but shouldn't this clear the value of the template and make your usecase work again?

oddlama: You have to read the upstream nixpkgs source I linked before. You will see that `restartIfChanged` is not an option in `serviceConfig` and thus cannot be used like you are using it. I believe you should use `serviceConfig."X-RestartIfChanged"` instead.

SuperSandro2000: good point. Not having a good one with this PR 😂

Mohswalm545:
You have to read the upstream nixpkgs source

oddlama:
Thanks @SuperSandro2000, this looks a lot more sound. I successfully tested this after a minor modification:

```diff
@@ -216,7 +216,7 @@ in
       };
       "microvm@${name}" = {
         # restartIfChanged is opt-out, so we have to include the definition unconditionally
-        inherit (microvmConfig) restartIfChanged;
+        serviceConfig.X-RestartIfChanged = [ "" microvmConfig.restartIfChanged ];
         # If the given declarative microvm wants to be restarted on change,
         # We have to make sure this service group is restarted. To make sure
         # that this service is also changed when the microvm configuration changes,
```

astro:
Are you sure that `""` makes sense?

oddlama:
Yes, the "" translates to an empty key `X-RestartIfChanged=` resetting any previous definitions (it's a systemd thing)

SuperSandro2000:
amended, thanks

oddlama (approved):

astro:
Thank you, guys.

If you feel that this could benefit from some explanatory comments, I'll merge them.

PR #113, opened 2023-06-30
boot-disk: Evaluate libguestfs-appliance only when needed

mikatammi:
Only evaluate libguestfs-appliance as part of the boot-disk when needed. This is to enable builds on aarch64-linux platforms against nixos-unstable.


astro:
Could you please check if this is still required with https://github.com/astro/microvm.nix/commit/1789270c96d4e8d324289c042614b50d9ebb2af5?

Issue #114, opened 2023-07-02
microvms sometimes fail to start, possible race condition


oddlama:
In around one out of 10 cases, a microvm may fail to come up. There seems to be a race condition somewhere, so a restart often fixes it. The unfortunate thing is that the vm tries to enter a emergency shell by default, meaning the corresponding `microvm@<instance>.service` is still running on the host, which makes it impossible to detect this failure without inspecting logs.

I'm not entirely sure yet what causes this exactly, but for some reason the virtiofs mount doesn't detect the tags even though the virtiofsd server tells me that a client has connected successfully.

```
Jul 02 01:22:06 ward systemd[1]: Starting VirtioFS daemons for MicroVM 'loki'...
Jul 02 01:22:06 ward systemd[1]: Started VirtioFS daemons for MicroVM 'loki'.
Jul 02 01:22:06 ward microvm-virtiofsd@loki[2714343]: [2023-07-01T23:22:06Z INFO  virtiofsd] Waiting for vhost-user socket connection...
Jul 02 01:22:06 ward microvm-virtiofsd@loki[2714341]: [2023-07-01T23:22:06Z INFO  virtiofsd] Waiting for vhost-user socket connection...
Jul 02 01:22:06 ward microvm-virtiofsd@loki[2714344]: [2023-07-01T23:22:06Z INFO  virtiofsd] Waiting for vhost-user socket connection...
Jul 02 01:22:06 ward microvm-virtiofsd@loki[2714343]: [2023-07-01T23:22:06Z INFO  virtiofsd] Client connected, servicing requests
Jul 02 01:22:06 ward microvm-virtiofsd@loki[2714344]: [2023-07-01T23:22:06Z INFO  virtiofsd] Client connected, servicing requests
Jul 02 01:22:06 ward microvm-virtiofsd@loki[2714341]: [2023-07-01T23:22:06Z INFO  virtiofsd] Client connected, servicing requests
```

Client connects at `01:22:06` without errors. Below is the tty output of the microvm (somewhat interleaved apparently due to systemd parallelism unfortunately), which shows that for some reason the virtofs doesn't see the tags. `virtio-fs: tag <ro-store> not found`. I didn't do anything special with the shares, it's basically just like the examples for forwarding `/nix` as `.ro-store`, plus two extra shares for impermanence.

[...]

PR #117, opened 2023-07-28
fix: always wait for module loading for critical file systems and virtiofs

oddlama:
- This is a fix that closes https://github.com/astro/microvm.nix/issues/114

Since a fix upstream in nixpkgs seems to be non-trivial (the necessary modules are not in a direct dependency chain like with 9p), I would add these barriers to ensure that filesystem modules are always available.


astro:
Thank you!

PR #122, opened 2023-08-03
fix: stratovirt link

vedkothavade:
add correct stratovirt link


astro:
Oops, thank you!

vedkothavade:
np! this project looks very cool, thanks for making it 😄

PR #126, opened 2023-08-17
README: propose commercial support

zimbatm:
Provide a path for companies that want to spend money to improve the
project.


06kellyjac:
I'm not sure if this CI will end up skipping once approved or something but it probably shouldn't run if docs such as README are changed considering there's 112 checks :sweat_smile:

astro:
:heart:

PR #136, opened 2023-09-12
Fix issue related to romfile= argument

mikatammi:
Fixes the `Property 'virtio-net-device.romfile' not found` error, if running `-M microvm` -kind of configuration on x86_64-linux.


astro:
Thank you!

PR #141, opened 2023-09-25
Update cloud hypervisor

phaer:
Hello,

This PR:

* Updates the vendored version of cloud-hypervisor from upstream (identical in nixpkgs and spectrums nixpkgs atm)
* ~Removes rust-hypervisor-firmware because it's [currently broken](https://github.com/NixOS/nixpkgs/issues/256957) in nixpkgs and I don't think we need it, as cloud-hypervisor can boot kernel + initrds directly and we don't seem to depend on bootloaders?~
* ~Maybe controversal: rename cloud-hypervisor-graphics to just cloud-hypervisor and use the same build for graphical and non-graphical hosts alike. Happy to remove this or make it opt-in if needed, see commit msg for justification.~

### DIFF
--- a/pkgs/spectrum-os/cloud-hypervisor/0001-build-use-local-vhost.patch
+++ b/pkgs/spectrum-os/cloud-hypervisor/0001-build-use-local-vhost.patch
@@ -1,4 +1,4 @@
-From 9ac46605a87746d2d3e5a46a75cde33a7f01d31c Mon Sep 17 00:00:00 2001
+From 553849a8c6fd835909f8d0fac52578cae120c4f1 Mon Sep 17 00:00:00 2001
 From: Alyssa Ross <alyssa.ross@unikie.com>
 Date: Wed, 28 Sep 2022 12:18:19 +0000
 Subject: [PATCH 1/2] build: use local vhost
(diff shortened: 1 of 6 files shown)


astro:
I am terribly sorry for the duplicate work but I had the `--initramfs` done myself already.

The cloud-hypervisor graphics patch has uncertain future. It's nice to support it while it's fresh but my primary concern is headless operation using the latest cloud-hypervisor release which has its best place in upstream nixpkgs.

phaer:
I've removed everything but the upgrade, because that might still be useful?

> I am terribly sorry for the duplicate work but I had the --initramfs done myself already.

No worries, good to have it :)

> The cloud-hypervisor graphics patch has uncertain future. It's nice to support it while it's fresh but my primary concern is headless operation using the latest cloud-hypervisor release which has its best place in upstream nixpkgs.

Ah, interesting - do you happen to have a link regarding the future of cloud-hypervisors graphic patchset? 

Just read https://spectrum-os.org/lists/hyperkitty/list/devel@spectrum-os.org/thread/NQ76ZY2ZKDVU7IWPYO5PIASBX27MJVKO/ and that had left the impression there was still active development in this area?

astro:
It is awesome that Alyssa updates the patchset. Yet it is not merged into upstream cloud-hypervisor and that will probably never happen. Therefore, it is going to lag behind upstream releases eventually.

Maybe help lobbying for acceptance in cloud-hypervisor? :shrug:

PR #142, opened 2023-09-26
remove mktuntap for crosvm

phaer:
The current crosvm runner uses mktuntap to pass fds of the network devices to a vm. It seems to silently ignore declared mac addresses as well as non tap-devices.

Current versions of crosvm in nixpkgs also support a --net flag where tap id & mac can be passed directly, so we use that and throw an error if non-tap devices were declared.

I believe mktuntap is not needed (anymore), so it's been removed here. Please let me know if it still provides value!


astro:
Passing as fd was required for macvtap interfaces. Maybe some individual hypervisors can open it themselves (by now)?

phaer:
I think the mkuntap invocation in the crosvm runner is the only one and that is, if i read correctly, behind a `lib.optionals (type == "tap")` guard, so it should not affect `macvtap` devices as they were seemingly unsupported with microvm.nix+crosvm?

Could look into adding support for them in this PR if needed, even though I don't use them personally. Passing the fds via the new-ish `--net tap-fd=$fd,mac=$mac` rather than the deprecated `--tap-fd` would also allow setting the mac :)

astro:
You are right. I remembered this wrongly.

Great that this is now possible with crosvm. Thank you!

PR #144, opened 2023-09-29
Remove dependency on linux-dev

lf-:
This dependency is not necessary, since qemu knows how to boot a bzImage directly.

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -39,7 +39,7 @@ let
             then "pci"
             else "device";
   kernelPath = {
-    x86_64-linux = "${kernel.dev}/vmlinux";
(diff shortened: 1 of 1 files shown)


astro:
Right, thank you!

Because that could be unified with the aarch64 case, I did the change myself.

lf-:
Oh yes it absolutely could. I think I was going to go that and then got distracted by being in a hurry. I also did a bunch of closure size and boot time work, I'll see about making a PR for that soon. 

(our use case is the Nix based challenge flakey-ci in maplectf, which is microvm.nix inside nix2container on k8s, which actually works so yay http://ctf.maplebacon.org)

astro:
Awesome

PR #145, opened 2023-09-30
Update faq.md systemd-tmpfiles

sbc64:
systemd-tmpfiles config has [7 columns](https://www.man7.org/linux/man-pages/man5/tmpfiles.d.5.html). This PR improves the faq example and avoid the following error:

`/etc/tmpfiles.d/00-nixos.conf:4: Invalid age '/var/lib/microvms/<vmname>/journal/<machineId>'.`

EDIT:

Thank you very much for your work on microvm.nix 🙏


astro:
Thank you taking care of that!

PR #146, opened 2023-10-01
nixos-modules: add closure size + startup time optimization

lf-:
This is based on my work on MapleCTF to run microvm.nix inside a docker
container (incidentally, an awesome microvm.nix use case) and have the
size not cause substantial issues.

Overall this saves about 700MB of closure size of a naive no-op VM
configuration at practically the sole cost of eating a qemu compile.

```
co/microvm.nix » nix path-info -sSh ./result1
/nix/store/ligbkxkl1hnz2pvj8d9dfic991zfc0s0-microvm-qemu-nixos     1.7K  756.3M

co/microvm.nix » nix path-info -sSh ./result-without
/nix/store/h218db586pc627ai96j5cq2gssbvaxk8-microvm-qemu-nixos     1.7K    1.4G
```

Without:
```
[root@nixos:~]# systemd-analyze time
Startup finished in 1.402s (kernel) + 8.830s (userspace) = 10.233s
multi-user.target reached after 8.830s in userspace.
```

With:
```
[root@nixos:~]# systemd-analyze time
Startup finished in 295ms (kernel) + 2.066s (initrd) + 2.471s (userspace) = 4.834s
multi-user.target reached after 2.466s in userspace.
```

This looks impressive, and perhaps it is, but there is a lot of speed to
still get out. The dominant factors in each system startup are:

Old:
- dhcpcd.service (7.706s)
- firewall.service (917ms)
- sshd.service (518ms)

New:

This group (in total, 3.1s ish in initrd; I have no idea why they are
this bad!):

- sys-devices-platform-serial8250-tty-ttyS1.device (3.087s)
- dev-ttyS1.device (3.087s)
- sys-devices-platform-serial8250-tty-ttyS2.device (3.085s)
- dev-ttyS2.device (3.085s) dev-ttyS0.device (3.084s)
- sys-devices-pnp0-00:02-tty-ttyS0.device (3.084s)
- sys-devices-platform-serial8250-tty-ttyS3.device (3.082s)
- dev-ttyS3.device (3.082s)
- dev-disk-by\x2dpath-pci\x2d0000:00:02.0.device (3.181s)
- sys-devices-pci0000:00-0000:00:02.0-virtio1-block-vda.device (3.181s)
- dev-disk-by\x2duuid-0d9f5dc5\x2d0d09\x2d4a74\x2d88a5\x2da0a5be825cc5.device (3.181s)
- dev-vda.device (3.181s)
- dev-disk-by\x2ddiskseq-1.device (3.180s)
- dev-disk-by\x2dpath-virtio\x2dpci\x2d0000:00:02.0.device (3.180s)
- systemd-fsck@dev-vda.service (12ms)
- sys-module-fuse.device (2.968s)

Then:
- systemd-udev-trigger.service (261ms)
- firewall.service (735ms) [!!!!!]
- systemd-networkd.service (273ms)
- sshd.service (511ms)

TL;DR: overall a huge improvement but it could have another 50% shaved
off, and I really have no idea why all that hardware init takes 3
seconds!

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -6,17 +6,33 @@
 let
   inherit (pkgs) lib system;
 
-  qemu =
-    if lib.any ({ bus, ... }: bus == "usb") microvmConfig.devices
-    then pkgs.qemu_kvm.overrideAttrs (oa: {
-      configureFlags = oa.configureFlags ++ [
-        "--enable-libusb"
-      ];
-      buildInputs = oa.buildInputs ++ (with pkgs; [
-        libusb
-      ]);
-    })
-    else pkgs.qemu_kvm;
+  enableLibusb = pkg: pkg.overrideAttrs (oa: {
+    configureFlags = oa.configureFlags ++ [
+      "--enable-libusb"
+    ];
+    buildInputs = oa.buildInputs ++ (with pkgs; [
+      libusb
+    ]);
+  });
+
+  minimizeQemuClosureSize = pkg: (pkg.override (oa: {
+    # standin for disabling everything guilike by hand
+    nixosTestRunner = if graphics.enable then oa.nixosTestRunner else true;
+    enableDocs = false;
+  })).overrideAttrs (oa: {
+    postFixup = ''
+      ${oa.postFixup or ""}
+      # This particular firmware causes 192mb of closure size
+      ${lib.optionalString (system != "aarch64-linux") "rm -rf $out/share/qemu/edk2-arm-*"}
+    '';
+  });
+
+  overrideQemu = x: lib.pipe x (
+    lib.optional (lib.any ({ bus, ... }: bus == "usb") microvmConfig.devices) enableLibusb
+    ++ lib.optional (microvmConfig.optimize.enable) minimizeQemuClosureSize
+  );
+
+  qemu = overrideQemu pkgs.qemu_kvm;
 
   inherit (microvmConfig) hostName vcpu mem balloonMem user interfaces shares socket forwardPorts devices graphics storeOnDisk kernel initrdPath storeDisk;
   inherit (microvmConfig.qemu) extraArgs;
--- a/nixos-modules/microvm/default.nix
+++ b/nixos-modules/microvm/default.nix
@@ -16,6 +16,7 @@ in
     ./system.nix
     ./mounts.nix
     ./graphics.nix
+    ./optimization.nix
   ];
 
   config = {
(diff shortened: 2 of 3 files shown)


lf-:
You can get the systemd-analyze plot outputs here: https://gist.github.com/lf-/961f8f6be2de4f08f33a0cae505fd4c8

lf-:
(systemd bug this works around: https://github.com/systemd/systemd/issues/29388)

lf-:
oh, the nonsense device times are actually just, the entire initrd basically. https://github.com/systemd/systemd/issues/29010

lib/runners/cloud-hypervisor.nix
```diff
@@ -65,6 +65,9 @@ let
   };
 
 in {
+  hypervisor = "crosvm";
```
astro: This is not the `crosvm` file ;-)

lib/runners/cloud-hypervisor.nix
```diff
@@ -65,6 +65,9 @@ let
...
+  executable = pkgs.crosvm;
```
astro: You called this `package` elsewhere. I'm not sure if this is a very descriptive name, either.

astro (changes requested):
I really like the `microvm.optimize` idea a lot.

Could you drop the `passthru` commit? Instead, just tell me what you need, and I'm going to add that together with docs within a few days...

lib/runners/cloud-hypervisor.nix
lf-: gah i renamed it to package because i thought executable was a rubbish name and messed this bit up. 

I can just drop the commit, yeah. What I want is to be able to nix build something and get the modified qemu that will actually be used. Currently it only exists as an interpolation in a string so it's annoying to build the qemu to look at it.

lf-:
Surplus commit has been dropped.

lib/runners/cloud-hypervisor.nix
astro: I added `passthru.hypervisor` in f92c945. Do you still need something along `hypervisorExe`? For debugging purposes only?

lf-: That change is buggy since it gives what the user configured in all cases, rather than what the script is generated for:

```
Loading installable 'nixosConfigurations.boxMicro.config.microvm'...
Added 28 variables.
nix-repl> runner
runner
nix-repl> runner.
runner.cloud-hypervisor  runner.crosvm            runner.firecracker       runner.kvmtool           runner.qemu              runner.stratovirt
nix-repl> runner.crosvm.hypervisor
"qemu"
```

Also, yeah, I would like some attribute that is the hypervisor derivation.

astro: Fixed in c28a0e0

Issue #131, opened 2023-08-31
Support specifying vsock connections


Sohalt:
It would be nice to be able to specify vsock connections for VMs, e.g. have cloud-hypervisor launch with the `--vsock /some/path` option.

astro:
Good idea!

Do you think configuring just the socket path will be enough? What is your primary use-case? I was thinking about running [waypipe](https://gitlab.freedesktop.org/mstoeckl/waypipe/) through vsocks.

Sohalt:
I wanted to have a way for a custom service to communicate between vm and host without using up ports on the host. For that specifying the path should be enough.

astro:
What about the CID?

astro:
Depending on the VMM, VSOCK are terminated differently:

* qemu, kvmtool, and crosvm support host AF_VSOCK which feels first-class. I added support for these three. If you want to translate into other socket domains use socat.
* cloud-hypervisor translates VSOCK connections to Unix Domain sockets, with the port interpolated into their paths.
* firecracker translates VSOCK connections to Unix Domain sockets but exposes a proxy protocol to the host so that the port can be passed.

Ramblurr:
@Sohalt Astro implemented qemu vsock support back in 2023, and I've just now prepared an implementation PR for cloud-hypervisor  that you may want to play with: https://github.com/astro/microvm.nix/pull/379

PR #147, opened 2023-10-03
crosvm: sync gpu params with cloud-hypervisor

phaer:
Note that "hidden = true" for the display doesn't work as intended for me, needed to patch crosvm to actually get rid of the display showing tty1. But that's true for cloud-hypervisor as well.

### DIFF
--- a/lib/runners/crosvm.nix
+++ b/lib/runners/crosvm.nix
@@ -31,7 +31,14 @@ let
     aarch64-linux = "${kernel.out}/${pkgs.stdenv.hostPlatform.linux-kernel.target}";
   }.${system};
 
+  gpuParams = {
+    context-types = "virgl:virgl2:cross-domain";
+    egl = true;
+    vulkan = true;
+  };
+
 in {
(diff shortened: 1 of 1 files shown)


astro:
That `hidden` parameter didn't work for me either.

Great that you're improving the graphical use-case! Since #131 I have that idea of trying waypipe which got vsock support 2 days ago: https://gitlab.freedesktop.org/mstoeckl/waypipe/-/merge_requests/16

phaer:
> That `hidden` parameter didn't work for me either.

Good to know, thanks!

> Great that you're improving the graphical use-case! Since #131 I have that idea of trying waypipe which got vsock support 2 days ago: https://gitlab.freedesktop.org/mstoeckl/waypipe/-/merge_requests/16

Ah, I had taken a quick look into waypipe but hadn't seen that PR so far. Might be interesting for many use cases, but I think for desktop vms virtio-gpu with crosvm might be more desirable as it supports opengl et al?

PR #149, opened 2023-10-05
Set type of flake option to string, correctly evaluate it when provided

### LINKED ISSUE #148: Flake option is a path but expects a set later on
The flake option is [expected to be a path](https://github.com/astro/microvm.nix/blob/abd63123e2b2dbc34d1ac38a73578b27ec9ef342/nixos-modules/host.nix#L74), but when passing a path I get the following error:

```
error: value is a path while a set was expected

       at /nix/store/wjwcshj9w20wdfpn4z1dc1sg8h539ix5-source/nixos-modules/host.nix:177:28:

          176|         guestConfig = if isFlake
          177|                       then flake.nixosConfigurations.${name}.config
             |                            ^
          178|                       else microvmConfig.config.config;
(use '--show-trace' to show detailed location information)

```

https://github.com/astro/microvm.nix/blob/abd63123e2b2dbc34d1ac38a73578b27ec9ef342/nixos-modules/host.nix#L177

b-m-f:
Closes #148

### DIFF
--- a/nixos-modules/host.nix
+++ b/nixos-modules/host.nix
@@ -71,7 +71,7 @@ in
 
           flake = mkOption {
             description = "Source flake for declarative build";
-            type = nullOr path;
+            type = nullOr str;
             default = null;
           };
 
@@ -174,7 +174,7 @@ in
         inherit (microvmConfig) flake updateFlake;
         isFlake = flake != null;
         guestConfig = if isFlake
-                      then flake.nixosConfigurations.${name}.config
+                      then (builtins.getFlake flake).nixosConfigurations.${name}.config
                       else microvmConfig.config.config;
         runner = guestConfig.microvm.declaredRunner;
       in


astro:
The primary use-case for this option was to pass a flake that is `self` or a flake input, not a string. Can we support both?

astro:
When testing this with a flakeref string as `flake`, which is your use-case, I get `error: cannot call 'getFlake' on unlocked flake reference 'git+https://gitea.dev/user/nix-config.git', at «none»:0 (use --impure to override)`

With `--impure` I get `error: infinite recursion encountered` :frowning_face:

b-m-f:
Thanks for giving it a try.

For the  "impurity" I found https://github.com/NixOS/nix/issues/6124#issuecomment-1073053270 .


Now for the recursion.

In my setup I currently have 2 gitrepos:

- git+https://my-git/b-m-f/host
- git+https://my-git/b-m-f/dev-microvm

Have you maybe tried to reference flake in the same way as self?

Since the usage does not seem clear, maybe a new option could be added? How about `remoteRef`?

b-m-f:
@astro should I close this as 'feature not desired'?

astro:
I haven't yet taken more time to investigate this myself.

b-m-f:
@astro Alright. There is absolutely no urgency on my side.
I won't have free time for a while either, but maybe I can get back to this with a reducible project at some point.

SuperSandro2000:
Does https://github.com/NixOS/nix/blob/master/doc/manual/rl-next/getflake-path.md help us resolve this?

Issue #143, opened 2023-09-28
crosvm: add macvtap support


astro:
Follow-up to #142

PR #150, opened 2023-10-07
Add macvtap support for crossvm

pinkisemils:
This solves #143, had this code lying around for a while, but finally got around to pushing it. It's not a big change, and I've only verified that it works in my own limited test case.


astro:
Thanks for taking the initiative.

I had to add use of `macvtapFds` to make the behaviour consistent with the other hypervisors. Better have a look if that's still fitting your use-case!

Issue #152, opened 2023-10-14
Possible early termination in microvm-shutdown?


JohannesEbke:
I'm currently chasing an issue where a microvm seems to be terminated with prejudice (or alternatively loses access to the console and virtio filesystems) during shutdown.

I have used bcc exitsnoop to try to time the different exit times of the termination process, and have recorded this list of things that terminate when I type "systemctl stop microvm@survey.service" (The VM "survey" has been running for ~18seconds):

 ```TIME-CEST    PCOMM            PID     PPID    TID     AGE(s)  EXIT_CODE 
23:29:17.494 curl             21042   21041   21042   0.01    0
23:29:17.996 socat            21043   21041   21043   0.50    0
23:29:17.996 microvm-shutdow  21041   1       21041   0.53    0
23:29:17.998 (sd-close)       21059   1       21059   0.00    0
23:29:17.998 virtiofsd        20922   20909   20922   18.85   0
23:29:17.999 virtiofsd        20956   20933   20956   18.84   0
23:29:17.999 virtiofsd        20944   20914   20944   18.85   0
23:29:17.999 virtiofsd        20914   1       20914   18.86   0
23:29:18.000 virtiofsd        20921   20904   20921   18.86   0
23:29:18.000 virtiofsd        20904   1       20904   18.87   0
23:29:18.026 kvm-nx-lpage-re  20988   2       20988   18.85   0
23:29:18.027 kvm              20987   2       20987   18.85   0
23:29:18.060 cloud-hyperviso  20970   1       20970   18.89   0
23:29:18.061 (sd-close)       21074   1       21074   0.00    0
23:29:18.064 virtiofsd        20933   1       20933   18.92   0
23:29:18.075 rm               21076   21075   21076   0.00    0
23:29:18.075 microvm_-post-s  21075   1       21075   0.01    0
23:29:18.093 (sd-close)       21077   1       21077   0.00    0
23:29:18.095 virtiofsd        20909   1       20909   18.96   code 1
23:29:18.096 systemd-tty-ask  21040   21039   21040   0.63    0
23:29:18.097 systemctl        21039   15099   21039   0.64    0
```

Looking at the microvm-shutdown script, it seems to me that it is intended to wait for the whole thing to terminate, but either A) the socket is closed early or B) there is a bug in the socat call. Anyway this seems to lead systemd to an immediate termination of the virtiofsds, which could lead to the observed behaviour.

I'd be ready to post more information and also run more debugging.

[...]

PR #153, opened 2023-10-16
cloud-hypervisor: resolve issue #152 using waitpid

JohannesEbke:
The systemd unit ExecStop script for cloud-hypervisor used to wait for machine termination by detecting the closing of the cloud-hypervisor API socket. However, it seems that this socket is actually closed before the machine is safely shut down.

Therefore, this patch replaces the listen on the socket with a waitpid on the main PID, which is the cloud-hypervisor process. This process only terminates after the machine has properly shut down.

More information and experiment results are in #152

### DIFF
--- a/lib/runners/cloud-hypervisor.nix
+++ b/lib/runners/cloud-hypervisor.nix
@@ -151,8 +151,7 @@ in {
 
         api -X PUT http://localhost/api/v1/vm.power-button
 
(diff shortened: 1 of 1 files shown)


astro:
I use mostly cloud-hypervisor so I am directly affected by this. Thank you very much!

Issue #116, opened 2023-07-28
simplify simple-network.md


flokli:
Currently, `simple-network.md` describes the network setup for MicroVMs using a TAP interface, which is then attached to a bridge, and then lists some systemd-network config necessary on the host to configure the network for internet access.

It however can become even simpler, *by only creating the TAP interface, and letting the rule systemd-networkd ships with take care of providing client internet access out of the box, without any additional config necessary on the host* (other than systemd-networkd):

https://github.com/systemd/systemd/pull/15447 added a [80-vm-vt.network](https://github.com/systemd/systemd/pull/15447/files#diff-7f84efcbf57c0c78206ebcb424ba8daecb7f6b52422b79035e2f66217bc18793) file which automatically spins up a DHCP server and configures NAT for TAP interfaces starting with `vt-`. This has been in systemd since more than 3 years, so we can assume it to be present.

My proposal would be to extend/update the microvm.nix tooling to create TAP interfaces with that prefix, and then simplify these docs. We could still document how to manually explicitly configure networking, but it should probably also be on the TAP interface, for better isolation of individual VMs from each other.

Mic92:
@astro I can work on this, if you give me green light.

astro:
Good idea, go ahead

Mic92:
I played around with networkd's tap network unit but it wasn't working well.

1. Only ipv4 is supported:
  * It uses the `Address=0.0.0.0/24` to assign an ipv4 subnet. There is also `Address=::/64` for ipv6 however the dhcpv6 or router advertisment won't pick up subnets assigned like that.
2. Multiple subnets. Somehow I would after a while always end up with multiple subnets after a while on the interface instead of just one. This also sometimes broke the networking.

pwaller:
FYI, I had difficulty getting the networking working, following the docs.

The root cause was a difference between simple-network.md and interfaces.md:

https://github.com/astro/microvm.nix/blob/67bf176ff48141292a66d8bce541463ee49d0fb5/doc/src/interfaces.md#L11

https://github.com/astro/microvm.nix/blob/67bf176ff48141292a66d8bce541463ee49d0fb5/doc/src/simple-network.md#L53

The id in the former does not match the expression in the latter, so the networkctl does not bind the tap interface to the bridge and therefore ends up in the 'no-carrier' state. Took me a while to figure out what was going on.

Scanning through the docs, it wasn't so easy to put together a working configuration quickly.

astro:
[...]

PR #158, opened 2023-10-30
simple-network.md: Adapt to place microVMs directly on LAN

jpds:
This is much simpler than some odd private bridge/port-forwarding.

Fixes: #116.


astro:
That will work if you operate your host in your private LAN that usually has 253 IP addresses at your disposal. One other common use-case is running MicroVMs on a rented server with one public IPv4 address, requiring NAT and port forwarding for the VMs.

Both use-cases are valid and common. I propose we split the network setup into two documents. Would you like to do that?

jpds:
@astro Done - this should be more useful for people that just want to try microVMs locally.

astro:
Thank you very much!

PR #163, opened 2023-11-17
Make Kernel Samepage Merging optional

KenMacD:
As MicroVMs work fine without KSM enabled, it should probably be optional. While KSM can improve a lot of workloads, it can harm others, and so this configuration should not conflict with a value explicitly set by the user.


astro:
Thank you

Issue #127, opened 2023-08-17
Cross-system flake question


ereslibre:
First of all, thank you for microvm.nix! I am starting to work with it; I am writing a flake that can be directly executed (without the need of the host module). Right now, my flake looks like this:

```nix
{
  description = "microvm.nix-test";

  inputs = {
    microvm = {
      url = "github:astro/microvm.nix";
      inputs.nixpkgs.follows = "nixpkgs";
    };
    nixpkgs.url = "nixpkgs/nixos-23.05";
  };

  outputs = {
    self,
    microvm,
    nixpkgs,
  }: {
    nixosConfigurations = let
      pkgs = nixpkgs.legacyPackages."x86_64-linux";
    in {
        generic-dev = nixpkgs.lib.nixosSystem {
          system = "x86_64-linux";
          modules = [
            microvm.nixosModules.microvm
            {
              microvm = {
                interfaces = [
                  {
                    type = "user";
                    id = "microvm-test";
                    mac = "02:00:00:00:00:01";
                  }
                ];
                hypervisor = "qemu";
              };
              networking.hostName = "nixity-vm";
              nix.settings.experimental-features = ["nix-command" "flakes"];
              environment.systemPackages = with pkgs; [ htop ];
              services.getty.autologinUser = "root";
              system.stateVersion = "23.05";
            }
          ];
        };
      };
    };
}
```

Running `nix run .#nixosConfigurations.generic-dev.config.microvm.declaredRunner` works perfectly and I'm presented with a logged-in session at the created machine.

My next step was to have this virtual machine `aarch64-linux`. The host is `x86_64-linux`. [From the documentation](https://astro.github.io/microvm.nix/declarative.html#fully-declarative) I understand that this would be possible, if I was using the host module:

```nix
# microvm refers to microvm.nixosModules
{ microvm, ... }: {
  imports = [ microvm.host ];
  microvm.vms = {
    my-microvm = {
      # The package set to use for the microvm. This also determines the microvm's architecture.
      # Defaults to the host system's package set if not given.
      pkgs = import nixpkgs { system = "aarch64-linux"; };

      # ...
    };

    # ...
  };

  # ...
}
```

But I don't see how can I achieve the same thing when using the flake directly as in the example. I assume I have to plug the right `pkgs` [here](https://github.com/astro/microvm.nix/blob/a291d324915f26d1fd86443bd486089099e8b541/nixos-modules/microvm/default.nix#L1), but I am having no success doing so.

[...]

PR #165, opened 2023-12-07
feat: allow to specify a CPU to emulate with a cross-system

ereslibre:
Implements: https://github.com/astro/microvm.nix/issues/127

---

By setting a module of the form:

```
modules = [
  { nixpkgs.crossSystem.config = "aarch64-unknown-linux-gnu"; }
]
```

On the `lib.nixosSystem` call, and a specific CPU on the `nixosModules.microvm` call, such as:

```
microvm = {
  cpu = "cortex-a53";
}
```

Allow a different CPU to be emulated by using qemu's Tiny Code Generator (TCG). While this will have a performance impact on the guest, it allows to experiment cross-system derivations on an emulated guest.

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -32,24 +32,40 @@ let
     ++ lib.optional microvmConfig.optimize.enable minimizeQemuClosureSize
   );
 
-  qemu = overrideQemu pkgs.qemu_kvm;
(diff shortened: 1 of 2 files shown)


ereslibre:
Thanks to you @astro! In the following days I expect to document it explicitly in the projects documentation. I would also like to add at least a smoke/e2e test for the feature if you are fine with it.

PR #172, opened 2023-12-21
fix: don't set `neededForBoot = false` if not necessary

oddlama:
This PR just removes unnecessary definitions of `neededForBoot = false`, which are currently set on all normal volumes added by users. Removing these makes it possible to set `neededForBoot = true` for some of them (e.g. when using impermanence) without requiring use of `lib.mkForce true` every time.

This removes unnecessary conflict messages and makes it clear that the `neededForBoot` property is _not_ set to `false` explicitly by microvm.nix, because in reality just the `neededForBoot = true` values are those that are serve a purpose.


astro:
Thank you!

PR #174, opened 2023-12-22
Document CPU emulation with the qemu hypervisor

ereslibre:
Documents https://github.com/astro/microvm.nix/pull/165


astro:
Thank you very much!

PR #177, opened 2023-12-30
qemu: Fix cross-compilation

mikatammi:
Use microvmConfig.cpu == null to check whether we should be doing virtualization with KVM or emulation of the CPU, instead of checking pkgs.buildPlatform == pkgs.hostPlatform. The latter kind of check causes microvm not start, if the system running microvms has been cross-compiled for example from x86_64-linux to aarch64-linux.

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -35,8 +35,8 @@ let
     ++ lib.optional microvmConfig.optimize.enable minimizeQemuClosureSize
   );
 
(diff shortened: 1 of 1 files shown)


astro:
@ereslibre Please review

ereslibre (approved):
Thank you @mikatammi. Certainly this is more robust, didn't consider the cross-compiled microvm.nix was an option!

astro:
Thanks!

PR #180, opened 2024-01-06
Change microvm.host.enable to be false by default

SuperSandro2000:
Just importing a module shouldn't change things. 
This likely requires other changes.

### DIFF
--- a/nixos-modules/host.nix
+++ b/nixos-modules/host.nix
@@ -11,7 +11,7 @@ in
   options.microvm = with lib; {
     host.enable = mkOption {
       type = types.bool;
-      default = true;
+      default = false;
       description = ''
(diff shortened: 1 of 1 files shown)


astro:
As the host module is to be imported implicitly, this would break things for many users.

I am aware of the troubles with conditional module imports. I'm open for this if there's a compelling reason.

SuperSandro2000:
I did this because I explicitly needed to disable the module when building an iso where it isn't useful and just takes up space. 

https://gitea.c3d2.de/c3d2/nix-config/src/commit/e9a4c3de115b57fd8188d659c6c047be3cde9967/modules/microvm-host.nix#L26

astro:
On the other hand, the host module won't get in the way but provides the microvm command.

PR #187, opened 2024-01-18
fix: wait for the correct mount before attempting to create an overlayfs (closes #186)

### LINKED ISSUE #186: Futher occasional boot failures due to module load order
While I though I fixed the boot order issues from https://github.com/astro/microvm.nix/issues/114 in https://github.com/astro/microvm.nix/pull/117, it unfortunately looks like there are still some left. In the past days I've had an increasing amount of issues that prevent a particular microvm from coming up, due to overlayfs not being available.

Sometimes microvms with a writable store overlay won't come up because the overlayfs lower dir is not available. (The message could be interpreted as saying that the overlayfs module is missing, but this is [not the case](https://stackoverflow.com/questions/43265182/overlayfs-special-device-overlay-doesnt-exist))

```
Jan 18 14:04:36 sire-paperless mount[136]: mount: /sysroot/nix/store: special device overlay does not exist.
Jan 18 14:04:36 sire-paperless mount[136]:        dmesg(1) may have more information after failed mount system call.
Jan 18 14:04:36 sire-paperless kernel: overlayfs: failed to resolve '/sysroot/nix/.ro-store': -2
```

So for some reason /sysroot/nix/store is mounted before /sysroot/nix/.ro-store is available, which should never happen because it looks like there is already a dependency here:

https://github.com/astro/microvm.nix/blob/088ba565537eaef1041a87be5a44ca0daa4e1908/nixos-modules/microvm/mounts.nix#L142

But as it turns out, this is missing a `/sysroot` prefix. Will do a quick PR.

I'm actually quite curious how nobody else seems to be running into these issues, since they should affect everyone using a writable store overlay if I'm not mistaken... Guess I'm just a bit lucky. :P

oddlama:
See #186. Also adds a requires on module loading to 9p shares which can also be affected by module load order.

### DIFF
--- a/nixos-modules/microvm/mounts.nix
+++ b/nixos-modules/microvm/mounts.nix
@@ -38,7 +38,7 @@ lib.mkIf config.microvm.guest.enable {
       "/nix/store" = {
         device = roStoreDisk;
         fsType = storeDiskType;
-        options = [ "x-systemd.after=systemd-modules-load.service" ];
+        options = [ "x-systemd.requires=systemd-modules-load.service" ];
         neededForBoot = true;
       };
     }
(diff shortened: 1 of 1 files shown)


astro:
Awesome find. Thank you!

PR #194, opened 2024-01-24
qemu: add option microvm.qemu.serialConsole

jpruiz84:
For some applications like UART passthrough we want to disable the default qemu virtual serial (-serial chardev:stdio). With this flag in false we remove this virtual device and the ttyAMA0/S0 kernel console.


astro:
Thank you very much!

PR #195, opened 2024-01-25
Use new packages.${system}.default convention

thkoch2001:
The switch from defaultPackage to packages.${system}.default happened in https://github.com/NixOS/nix/issues/5532.

Additionally I removed one layer:
self.nixosConfigurations.my-microvm.config.microvm.declaredRunner; self.nixosConfiguration.config.microvm.declaredRunner;

The latter should be easier to understand for new users.


thkoch2001:
Sorry, while waiting for the checks to pass I learned that `nixosConfigurations` is a predefined output for flakes and thus it might actually be confusing to have a `nixosConfiguration` (without plural s). On the other hand I don't believe that a microvm in nixosConfigurations is ever useful outside of the context of microvm?

The diff view can be simplified by surpressing whitespace diff.

astro:
Happy to catch up with `packages.${system}.default`.

There is no `nixosConfiguration` in the [Flakes schema](https://nixos.wiki/wiki/Flakes#Output_schema), only `nixosConfigurations`, sorry.

flake-template/flake.nix
```diff
@@ -8,42 +8,36 @@
...
-                shares = [ {
-                  # use "virtiofs" for MicroVMs that are started by systemd
-                  proto = "9p";
-                  tag = "ro-store";
-                  # a host's /nix/store will be picked up so that no
-                  # squashfs/erofs will be built for it.
-                  source = "/nix/store";
-                  mountPoint = "/nix/.ro-store";
-                } ];
+      nixosConfiguration = nixpkgs.lib.nixosSystem {
```
mikatammi: Is this right? Shouldn't it be `nixosConfigurations.my-microvm = ` ?

astro: No, yes. This PR needs changes.

@thkoch2001 Are you going to do that?

thkoch2001:
Sorry for the double-push. I reduced the commit now to the trivial one-line change.

In my own local flake however I removed nixosConfigurations entirely from the flake:

```
  outputs = { self, nixpkgs, microvm }:
    let
      system = "x86_64-linux";
      nixosConfiguration = nixpkgs.lib.nixosSystem {
          inherit system;
          modules = [
            microvm.nixosModules.microvm
[...]
        };
    in {
      packages.${system}.default = nixosConfiguration.config.microvm.declaredRunner;
    };
```

astro:
Thank you!

PR #198, opened 2024-02-02
Set microvm hostname as process name in runner

c0deaddict:
Having multiple microvm's running on my server it's always hard to see in `top` which VM is taking up resources: all processes are named `cloud-hypervisor`. This patch renames the process `/nix/store/$hash-cloud-hypervisor` (or whatever hypervisor is used) to `microvm-$hostname` making it possible to distinguish the running VM's. I'm not sure if other people also find this desirable, if not, maybe this can be added behing a toggle flag?

PS. it only shows the microvm name in `top` if the full command line is shown (eg. pressing `c`).


astro:
I like this a lot. Thank you!

I guess adding a default-on option to opt out of it should make everyone happy.

PR #204, opened 2024-02-18
Add docker/systemd-networkd doc snippet

iofq:
I was troubleshooting this issue for a while, found the solution, and thought it'd be helpful to add to the docs. 

Essentially, the config laid out under "Simple Networking" breaks docker networking within the VMs, because it matches all `ether` interfaces, which includes the bridged `veth*` interfaces docker creates and uses for each container. Unmanaging these fixes the issue, as found in https://github.com/systemd/systemd/issues/28626.

### DIFF
--- a/doc/src/simple-network.md
+++ b/doc/src/simple-network.md
@@ -83,6 +83,23 @@ systemd.network.networks."20-lan" = {
 };
 ```
 
+## Docker and systemd-network
+
+If you use the above `systemd.network` bridge config and wish to run
+Docker containers inside your microvms using `virtualisation.docker`,
+you may need to add the following snippet to stop `systemd-networkd` from
+managing the bridged `veth*` interfaces Docker creates for each container.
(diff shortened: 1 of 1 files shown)


astro:
That's certainly helpful for docker users.

What do you think about replacing `matchConfig.Type = "ether";` with `matchConfig.Name = "e*";` in `20-lan.network` instead? Would that be enough?

iofq:
That would also work as long as the interface name is reliably set within the VMs. Maybe we can suggest `networking.usePredictableInterfaceNames = false;` in each VM config so it's always set to `eth0` and then just specify that.

I suppose I'd prefer to only affect the docker networks when needed rather than potentially causing another obscure issue by wildcarding `e*`

astro:
> I suppose I'd prefer to only affect the docker networks when needed rather than potentially causing another obscure issue by wildcarding `e*`

I understand. Thank you for your contribution!

PR #207, opened 2024-02-27
Add `securityModel` option to shared directories

### LINKED ISSUE #206: Add `securityModel` option to 9P shares for QEMU
Hello! I would like to thank you first for your project, which I find very friendly for a virtualization newbie like myself.

I am trying to set up a VM with a share to be mounted in `/home/myuser`. But in order to have this share as a usable home folder for `myuser`, I need to change its permissions. I noticed that all shares are mounted in `qemu` using the `none` security model, which automatically mounts all the shares as `root` (let me know if I am mistaken about this).

I feel for my use-case having `mapped` would be better.

So I forked the repo and added one :) . I would like to know if I did well or there is perhaps an easier way of doing this.

rmgaray:
Fix #206 .

### DIFF
--- a/nixos-modules/microvm/options.nix
+++ b/nixos-modules/microvm/options.nix
@@ -276,6 +276,11 @@ in
             type = nonEmptyStr;
             description = "Path to shared directory tree";
           };
+          securityModel = mkOption {
+            type = nullOr str;
+            default = "none";
+            description = "What security model to use for the shared. Default: none.";
+          };
           mountPoint = mkOption {
             type = path;
             description = "Where to mount the share inside the container";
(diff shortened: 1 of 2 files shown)


nixos-modules/microvm/options.nix
```diff
@@ -276,6 +276,11 @@ in
             type = nonEmptyStr;
             description = "Path to shared directory tree";
           };
+          securityModel = mkOption {
+            type = nullOr str;
```
astro: There are four possible values, an `enum` would be good fit.

astro:
Thank you!

Issue #212, opened 2024-03-07
qemu crashes when running a microvm


999eagle:
Every microvm instantly crashes with the following output from qemu:
```
/build/qemu-8.2.1/include/hw/i386/pc.h:144:PC_MACHINE: Object 0x5f9b9c9eb010 is not an instance of type generic-pc-machine
[1]    40361 IOT instruction (core dumped)
```

This is an upstream issue of qemu itself (see https://gitlab.com/qemu-project/qemu/-/issues/2142) when running `qemu-system-x86_64 -M microvm,accel=kvm -cpu host`, which `microvm.nix` uses on `x86_64-linux`. It can be reproduced both with the completely unchanged flake template from this repo and when using `qemu` from nixpkgs directly.
I'm currently working around this by patching the `microvm-run` script to use `pc` as the machine type instead of `microvm` which I'm fine with, but this could be a problem for others.

astro:
Unfortunately, none of my current computers have SGX.

As long as this is a problem with latest qemu version I'm going to take any PRs that work around this issue.

PR #213, opened 2024-03-08
qemu: fix crash when using sgx

999eagle:
This is the bare-minimum fix to make sure the qemu runner doesn't crash on hosts with SGX available. I'm open for ideas on how to make this configurable if that's necessary.

Fixes: #212
Upstream issue: https://gitlab.com/qemu-project/qemu/-/issues/2142


astro:
Looks promising. Thank you!

PR #214, opened 2024-03-11
runner: fixed cross-compilation

jpruiz84:
When running in aarch64 a x86 compiled version, the result microvm-run script uses x86 bash runtimeShell. Fixed by removing the buildPackages from runtimeShell.


astro:
I fear that is going to conflict with the qemu-tcg use-case.

@mikatammi @ereslibre Please review.

ereslibre (approved):
Can confirm this works as expected with qemu-tcg too.

PR #217, opened 2024-03-19
New Example: Host with predictable guests

RobWalt:
Disclaimer: I started looking into microvm.nix on last friday. I had no experience whatsoever with systemd-network before so please apologize:

- my lacking knowledge in general
- if I did something stupid

---

I was mainly looking for ways to virtualize services hosted on my servers without too much of a hassle. microvm.nix seemed fitting my use case and getting the first VM to run was amazingly easy. After that I felt a bit lost as a total networking novice and I came up with the approach in this PR which is basically just a remix of the `microvms-host.nix` example. My goal was to have the guest systems on predictable addresses so that I can use the hosts nginx to proxy them in the same way I'm used to doing it on my server.

I was also looking into configuring the private network completely statically, but I didn't get that to work. There's also this [advanced networking](https://astro.github.io/microvm.nix/advanced-network.html#port-forwarding) page in the handbook which seemed like it could've helped to solve my problem, but again I didn't get something to work as those fragments of code were to little for my level of networking knowledge.

I just feel like there should exist an boilerplatey example for unskilled people that does something like this so I figured I might as well contribute this akward solution I found. Feel free to ignore this PR and close it if it's an stupid idea.

Anyways: Thanks for creating the flake! I had a lot of fun playing around with it and learning more about nix and VMs! :)


astro:
Cool, I didn't know how to add static leases to networkd's DHCP server before.

On the other hand, it duplicates a lot of configuration from the other examples. I think it would be a good fit to merge it with `examples/microvms-host.nix`.

RobWalt:
Thanks for the feedback and I agree, merging the examples makes sense. I'll see what I can do!

RobWalt:
The feature is included in the existing example now, giving a static IP to the nested QEMU VM. It should always get the ip 10.0.13.37 now

astro:
Thank you!

PR #218, opened 2024-03-19
Allow overriding nixpkgs for VM

bouk:
This allows settings pkgs = null in the microvm config so it instantiates a new package set. This way you can configure nixpkgs inside the NixOS configuration.

Also add a 'nixpkgs' config option to allow overriding the nixpkgs path.

### DIFF
--- a/nixos-modules/host/default.nix
+++ b/nixos-modules/host/default.nix
@@ -28,33 +28,44 @@ in
             default = null;
             type = nullOr (lib.mkOptionType {
               name = "Toplevel NixOS config";
-              merge = loc: defs: (import "${toString config.pkgs.path}/nixos/lib/eval-config.nix" {
+              merge = loc: defs: (import "${config.nixpkgs}/nixos/lib/eval-config.nix" {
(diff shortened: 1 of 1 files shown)


astro:
What is the difference to setting `config.nixpkgs.pkgs`?

bouk:
I was trying to set some `nixpkgs` options inside the VM configuration but then I get this error:

```
Failed assertions:
- Your system configures nixpkgs with an externally created instance. `nixpkgs.config` options should be passed when creating the instance instead.
```

Because the `pkgs` option is passed to `eval-config.nix`. So, I want to override `pkgs` to be `null` so doesn't set this `pkgs`: https://github.com/NixOS/nixpkgs/blob/8c7a594e40bd450af9f029ff43525714831bd045/nixos/lib/eval-config.nix#L58-L61 

And then I need to get a `nixpkgs` reference somewhere to call `eval-config` so I had to add the nixpkgs argument for it.

astro:
Thank you!

PR #219, opened 2024-03-19
Forward NOTIFY_SOCKET over vsock for cloud-hypervisor

### LINKED ISSUE #205: Implement vmm.notify_socket
Systemd supports receiving a `vmm.notify_socket` credential so it can notify the host when the VM has finished loading: https://www.freedesktop.org/software/systemd/man/latest/systemd.system-credentials.html#vmm.notify_socket

If we implement this then the `microvm@` service can accurately reflect the VM status

bouk:
This will forward the systemd ready notification inside the VM so the
systemd service status reflects whether the VM has finished booting.

I had to use socat to forward the messages because the NOTIFY_SOCKET that systemd creates only accepts datagrams, but cloud-hypervisor only supports streaming unix sockets.

Fixes #205

### DIFF
--- a/lib/runner.nix
+++ b/lib/runner.nix
@@ -16,6 +16,7 @@ let
   };
 
   inherit (hypervisorConfig) command canShutdown shutdownCommand;
+  supportsNotifySocket = hypervisorConfig.supportsNotifySocket or false;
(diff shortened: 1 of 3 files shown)


astro:
Rebased as a439229a1af9e0fae3b3b21619c1983901a41bf7

Issue #35, opened 2022-04-24
default formater suggestion


GTrunSec:
we should have a default formater for nix instead of being human-readable by manual. The nixpkgs.alejandra might be a choice.

JayRovacsek:
Did a formatter get selected? Wanting to contribute if I can get a feature I'd like working but don't want to break current format conventions
cc @astro

astro:
Are there alternatives to alejandra? I remember to dislike its defaults.

JayRovacsek:
`nixfmt` or `nixpkgs-fmt` might be good options?

PR #220, opened 2024-03-19
add `formatter` flake output (`nixpkgs-fmt`)

RobWalt:
Implements the suggestion from here https://github.com/astro/microvm.nix/issues/35 and uses [`nixpkgs-fmt`](https://github.com/nix-community/nixpkgs-fmt). I opted for it since it looked the least weird. IMO `alejandra` has not enough spacing and `nixfmt` uses a bit too much.

Anyways, feel free to suggest using another formatter. I just want to have a consistent formatting here which I can apply right before committing so that my own auto-formatting isn't getting in the way 😅 

I separated adding the formatter and formatting the code here into two commits so we could optionally omit the latter and do the actual formatting later to prevent merge conflicts.

### DIFF
--- a/checks/default.nix
+++ b/checks/default.nix
@@ -3,171 +3,200 @@
 let
   variants = [
     # hypervisor
-    [ {
+    [{
       id = "qemu";
-      modules = [ {
+      modules = [{
         microvm.hypervisor = "qemu";
-      } ];
-    } {
-      id = "qemu-tcg";
-      modules = let
-        # Emulate a different guest system than the host one
-        guestSystem = if "${system}" == "x86_64-linux" then "aarch64-unknown-linux-gnu"
-          else "x86_64-linux";
-      in [
-        {
-          microvm = {
-            hypervisor = "qemu";
-            # Force the CPU to be something else than the current
-            # system, and thus, emulated with qemu's Tiny Code Generator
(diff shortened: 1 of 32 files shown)


checks/default.nix
```diff
@@ -3,185 +3,217 @@
...
-        microvm.volumes = [ {
-          image = "nix-store-overlay.img";
-          label = "nix-store";
-          mountPoint = config.microvm.writableStoreOverlay;
-          size = 128;
-        } ];
-      }) ];
-    } ]
+    }
+      {
```
astro, line 92: These curly braces being on different indentation levels illustrates well why I don't like any of the current formatters.

checks/default.nix
```diff
@@ -3,185 +3,217 @@
...
 import ./startup-shutdown.nix args //
 import ./shutdown-command.nix args //
 
-builtins.foldl' (result: hypervisor:
-  let
-    args = {
-      inherit self nixpkgs system hypervisor;
-    };
-  in
+builtins.foldl'
```
astro, line 207: I don't like this formatting.

astro (changes requested):
Thank you for your effort. I am terribly sorry but I am still not happy with the state of Nix formatting in 2024.

PR #225, opened 2024-04-02
WiP! Add solo5-spt support

ehmry:


astro:
Given that this does not build a much configurable OS nor does it use one of the existing VMMs, I don't want it here. Sincere apologies because it looks very interesting.

What I meant by sharing the [microvm conventions](https://astro.github.io/microvm.nix/conventions.html) was that you just have to replicate the following topology:

* flake.nix
  * `nixosConfigurations.*.config.microvm.declaredRunner`
    * bin/microvm-run
    * bin/microvm-shutdown
    * share/microvm/tap-interfaces

This should fit your use-case. Ops-wise it integrates with a microvm.nix host without any further setup. I am considering doing similar things for WFVM, system-manager, nix-openwrt-imagebuilder, and nixbsd if time allows.

ehmry:
Not a problem, in this case I might fork the repo and strip it down to just solo5.

astro:
Take a look at this: https://github.com/astro/microvm-solo5-spt :tada:

PR #231, opened 2024-04-15
Increased microvm default timeout stop to 2m30s.

jpds:
On a low-power i3 box, I keep having a thundering herd issue where my microVMs are _almost_ finished booting when systemd decides to restart them all again.


astro:
This also makes sense because it should be higher than the default timeout inside the VM. Thanks!

PR #232, opened 2024-04-18
Use xhci instead of ehci for USB

phaer:
Hello,

I've noticed that microvm.nix adds an EHCI controller for USB forwarding, while the documentation seems to recommend XHCI: https://www.qemu.org/docs/master/system/devices/usb.html#xhci-controller-support

Is there a reason for using EHCI? If so, I am happy to create an option instead.


astro:
There was no reason. Thank you!

Let's create an option once anyone actually needs it.

PR #243, opened 2024-05-23
Added support for btrfs filesystem.

makuru-dd:
I have added a few variables to the btrfs filesystem creation to optimize it, for small filesystems and x64 CPUs, changes in order:
- -M for small filesystems
- --csum xxhash for fast x64 PCs.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -32,6 +32,7 @@ rec {
     { image
     , label
     , size ? throw "Specify a size for volume ${image} or use autoCreate = false"
+    , mkfsExtraArgs
     , fsType ? defaultFsType
     , autoCreate ? true
     , ...
@@ -39,7 +40,7 @@ rec {
       (label != null && !autoCreate) "Volume is not automatically labeled unless autoCreate is true. Volume has to be labeled manually, otherwise it will not be identified"
       (let labelOption =
              if autoCreate then
-               (if builtins.elem fsType ["ext2" "ext3" "ext4" "xfs"] then "-L"
+               (if builtins.elem fsType ["ext2" "ext3" "ext4" "xfs" "btrfs"] then "-L"
                 else if fsType == "vfat" then "-n"
                 else (pkgs.lib.warnIf (label != null)
                   "Will not label volume ${label} with filesystem type ${fsType}. Open an issue on the microvm.nix project to request a fix."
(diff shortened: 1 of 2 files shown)


lib/default.nix
```diff
@@ -48,15 +49,18 @@ rec {
            labelArgument =
              if (labelOption != null && label != null) then "${labelOption} '${label}'"
              else "";
+            fsCreateExtraArgsParsedToMkfs =
+              if fsCreateExtraArgs != null then nixpkgs-lib.concatMapStringsSep " " (x: x) fsCreateExtraArgs
```
astro:
```suggestion
              if fsCreateExtraArgs != null
              then nixpkgs-lib.escapeShellArgs fsCreateExtraArgs
```

lib/default.nix
```diff
@@ -1,4 +1,4 @@
-{ nixpkgs-lib }:
+{ nixpkgs-lib, }:
```
astro:
```suggestion
{ nixpkgs-lib }:
```

nixos-modules/microvm/options.nix
```diff
@@ -198,10 +198,15 @@ in
             default = true;
             description = "Created image on host automatically before start?";
           };
+          fsCreateExtraArgs = mkOption {
+            type = listOf str;
+            default = [];
```
astro, line 203: Use the defaults from your first draft.

astro (changes requested):

lib/default.nix
```diff
@@ -48,15 +49,19 @@ rec {
            labelArgument =
              if (labelOption != null && label != null) then "${labelOption} '${label}'"
              else "";
+            fsCreateExtraArgsParsedToMkfs =
+              if fsCreateExtraArgs != null
+              then nixpkgs-lib.concatStringsSep " " fsCreateExtraArgs # I dunno why nixpkgs-lib.escapeShellArgs doesnt work with btrfs.
```
astro: Could you please show your problematic `fsCreateExtraArgs` value?

nixos-modules/microvm/options.nix
makuru-org: That wouldnt, work because of other FS then btrfs.

lib/default.nix
makuru-org: It doesn't want to see single as a parameter even though it exist. 
```
ERROR: unknown metadata profile  single
```

nixos-modules/microvm/options.nix
```diff
@@ -198,10 +198,15 @@ in
             default = true;
             description = "Created image on host automatically before start?";
           };
+          fsCreateExtraArgs = mkOption {
```
astro: I prefer `mkfsExtraArgs` for brevity.

astro (approved):

lib/default.nix
```diff
@@ -48,15 +49,19 @@ rec {
            labelArgument =
              if (labelOption != null && label != null) then "${labelOption} '${label}'"
              else "";
+             mkfsExtraArgsParsedToMkfs =
+              if fsCreateExtraArgs != null
+              then nixpkgs-lib.escapeShellArgs mkfsExtraArgs
+              else " ";
```
astro:
```suggestion
             mkfsExtraArgsString =
               if mkfsExtraArgs != null
               then nixpkgs-lib.escapeShellArgs mkfsExtraArgs
               else "";
```

astro (changes requested):

PR #248, opened 2024-06-06
QEMU balloon script abstraction improvement

slakkala:
Make the QEMU balloon script abstract away the difference between hypervisor balloon commands. The QMP balloon command expects the wanted memory size of the guest, so calculate the correct value from the balloon size given as argument.

Also some changes to reported balloon size; the changed signal is not emitted, if balloon size is not changed; and may report an old value.

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -342,11 +342,13 @@ in {
       VALUE=$(( (${toString (mem + balloonMem)} - $SIZE) * 1024 * 1024 ))
       SIZE=$( (
         ${writeQmp { execute = "qmp_capabilities"; }}
-        ${writeQmp { execute = "balloon"; arguments.value = 987; }}
-      ) | sed -e s/987/$VALUE/ | \
+        (${writeQmp { execute = "balloon"; arguments.value = 987; }}) | sed -e s/987/$VALUE/
+        ${pkgs.coreutils}/bin/sleep 2
+        ${writeQmp { execute = "query-balloon"; }}
+      ) | \
         ${pkgs.socat}/bin/socat STDIO UNIX:${socket},shut-none | \
         tail -n 1 | \
-        ${pkgs.jq}/bin/jq -r .data.actual \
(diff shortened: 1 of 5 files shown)


lib/runners/qemu.nix
```diff
@@ -339,16 +339,18 @@ in {
...
     if socket != null
     then ''
-      VALUE=$(( $SIZE * 1024 * 1024 ))
+      VALUE=$(( (${toString (mem + balloonMem)} - $SIZE) * 1024 * 1024 ))
       SIZE=$( (
         ${writeQmp { execute = "qmp_capabilities"; }}
-        ${writeQmp { execute = "balloon"; arguments.value = 987; }}
-      ) | sed -e s/987/$VALUE/ | \
+        (${writeQmp { execute = "balloon"; arguments.value = 987; }}) | sed -e s/987/$VALUE/
+        ${pkgs.coreutils}/bin/sleep 2
```
astro: The other VMMs also exhibit delayed effects. I think we should add the `sleep` to the other runners, too.

Also, the hardcoded `2` might be too low for loaded systems. Do you mind creating an option for it?

astro (changes requested):
Very cool that you're taking care of consistency.

I think we should add documentation regarding what the `$SIZE` argument actually means. A comment in the generated shell script would be enough.

lib/runners/qemu.nix
slakkala: Cloud-hypervisor didn't seem to require a delay, added delay to crosvm and made the delay configurable (still defaults to 2)

slakkala:
Added a description of `SIZE`. Also added balloon size query to cloud-hv's balloon script.

astro (approved):
I am happy with this. Thanks!

One last question before merging: do you really think it's intuitive to pass the amount of Megabytes to *remove* from the memory size? Personally, I would expect the number to be the size of the balloon memory available to the VM. Then again I don't use ballooning myself, so your opinion is all that matters to me.

mikatammi:
image

astro:
Deprecated by #343

PR #249, opened 2024-06-12
cloud-hypervisor: Use correct console on aarch64

slakkala:
Similarly to qemu, cloud-hypervisor also emulates a serial device that gets mapped to ttyAMA0 on aarch64, adjust kernel command line accordingly.


astro:
Thank you! :rainbow_flag:

PR #250, opened 2024-06-14
cloud-hypervisor: Remove hardcoded value of console

vunnyso:
We have kernelConsole value already defined but its not being used as argument. So removing hardcoded value and using kernelConsole value, which will fix issue console service crash issue on aarch64.


slakkala:
Oops, I accidentally missed this change in my PR, thanks @vunnyso for the fix.

astro:
Thanks

PR #251, opened 2024-06-18
networking/tap: make tap interfaces optional

MarcelCoding:


astro:
The less-code alternative would be defining:
```nix
systemd.services."microvm-tap-interfaces@%i.service".enable = false;
```

astro:
Btw, if the systemd services are in your way, you don't have to use `nixosModules.host`. You could build `...config.microvm.declaredRunner` and just run that `result/bin/microvm-run` in your environment.

PR #255, opened 2024-07-06
cloud-hypervisor: allow using a capability wrapper

SomeoneSerge:
I'd like to be able to have cloud-hypervisor create TAP devices without sudo. I'm still testing this, and I'm not sure how best to handle the graphics/non-graphics variants and versions, but the basic functionality of trying a "system" cloud-hypervisor and falling back to the store path is easy to achieve

### DIFF
--- a/lib/runners/cloud-hypervisor.nix
+++ b/lib/runners/cloud-hypervisor.nix
@@ -6,7 +6,69 @@
 let
   inherit (pkgs) lib system;
   inherit (microvmConfig) vcpu mem balloonMem user interfaces volumes shares socket devices hugepageMem graphics storeDisk storeOnDisk kernel initrdPath;
-  inherit (microvmConfig.cloud-hypervisor) extraArgs;
+  inherit (microvmConfig.cloud-hypervisor) extraArgs systemExecutable;
+
+  cloud-hypervisor =
+    if graphics.enable then pkgs.cloud-hypervisor-graphics else pkgs.cloud-hypervisor;
+
+  # Removes references to $out from the PATH to see if a security wrapper is
+  # installed
+  cloud-hypervisor-shim =
+    pkgs.runCommand "cloud-hypervisor-shim" { meta.mainProgram = "cloud-hypervisor"; }
+      (
+        ''
+          mkdir -p $out/bin
+          cat << EOF > "$out/bin/cloud-hypervisor"
+          #!${lib.getExe pkgs.bash}
+
+          set -e
+
+          # Inhibit the shim if it's in the PATH
+
+          new_PATH=
+          while IFS=: read -r -d ":" path ; do
+            if [[ "\$path" = *"$out"* ]] ; then
+              continue
+            fi
+            if [[ "\$new_PATH" == "" ]] ; then
+              new_PATH="\$path"
+            else
+              new_PATH="\$new_PATH:\$path"
+            fi
+          done <<< "\$PATH"
+          export PATH="\$new_PATH"
+
+          path="\$(type cloud-hypervisor 2>/dev/null)" \
+            || path=${lib.getExe cloud-hypervisor}
+          path="\''${path#cloud-hypervisor is }"
+
+          if [[ "\$path" == *"$out"* ]] ; then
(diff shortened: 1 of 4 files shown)


astro:
These are quite a number of changes to give more permissions to cloud-hypervisor. I wonder about alternatives.

Apart from tuntap the nixosModules.host systemd units also start virtiofsd. Maybe we could detect whether microvm-run was started through systemd, and if it was not, start virtiofsd instances and create the tap interface in a privileged tool script that has small concise functionality.

SomeoneSerge:
> These are quite a number of changes to give more permissions to cloud-hypervisor

Does feel a bit ad hoc, although conceptually I feel like "a fallback-capable wrapper that tries to use a system tool" should not be an uncommon scenario

> Maybe we could detect whether microvm-run was started through systemd

Tbh, I've been using this with `nix run .#...declaredRunner`:)

lib/runners/cloud-hypervisor.nix
```diff
@@ -6,7 +6,69 @@
...
+          expectedVersion="${lib.getVersion cloud-hypervisor}"
+          if [[ "\$foundVersion" != "\$expectedVersion"* ]]; then
+            echo "cloud-hypervisor-shim: sanity check failed, system \
+          cloud-hypervisor (\$foundVersion) doesn't match the expected version \
+          (\$expectedVersion). Try disabling \
+          microvm.cloud-hypervisor.systemExecutable.versionCheck or updating the \
+          host system" >&2
+            exit 1
+          fi
+        ''
```
SomeoneSerge, line 64: Btw I could delete this chunk. I thought I ran into a version mismatch issue, but it turned out to be a typo

nixos-modules/host/options.nix
```diff
@@ -10,6 +10,11 @@
       '';
     };
 
+    host.cloud-hypervisor.enable = lib.mkEnableOption "Install a cloud-hypervisor security wrapper" // {
+      default = true;
+    };
+    host.cloud-hypervisor.package = lib.mkPackageOption pkgs "cloud-hypervisor" { };
```
SomeoneSerge, line 16: In the hindsight, it might be sufficient to just expose ~~this~~ (EDIT: not this one, but make one for the guest description) one option. That'd have allowed me to implement the wrapper locally and configure `microvm.nix` to use it instead of the real thing. Even now I could do it it using an overlay, but re-evaluating Nixpkgs is a bit heavy handed for the task

SomeoneSerge:
Oh, I just noticed I get an error with qemu as well: `microvm@someone-0tier: -netdev tap,id=p-uvm-0tier,ifname=p-uvm-0tier,script=no,downscript=no: could not configure /dev/net/tun (p-uvm-0tier): Operation not permitted`.

So what is the general story for dynamic/imperative microVMs and security wrappers right now? A tangential question, getting far outside the scope of the PR, is what's the situation with managing virtiofsd services for imperative microVMs. I see there's an option to deploy a systemd template service for running these via supervisord, but that seems like a microvmCommand-specific solution...

SomeoneSerge:
>  I feel like "a fallback-capable wrapper that tries to use a system tool" should not be an uncommon scenario

Would you be more comfortable merging this if I tried to push a more generic wrapper implementation into Nixpkgs? 🙃

astro:
Not sure I have an idea what you mean by that. I really prefer having privileged tasks like host network setup outside of the VMM.

SomeoneSerge:
> Not sure I have an idea what you mean by that.

I meant that I can try and write a generic function for generating a wrapper that can try locating a security wrapper and fallback to the unprivileged store path. This would make the present PR look less clunky.

> I really prefer having privileged tasks like host network setup outside of the VMM.

Well this behaviour can be optional. In my case, and I appreciate yours is different and you may not wish to support mine, I seek to avoid having to maintain static configurations for my VMs in the hypervisor's nixos configuration, and maybe avoid the whole supervisord-inside-system business too

astro:
I guess you can still use microvms as packages, just start them differently (like with Nomad in [Skyflake](https://github.com/astro/skyflake)).

PR #256, opened 2024-07-09
Add the alioth VMM

astro:
Upstream PR: https://github.com/google/alioth/pull/72

Test with:
```
nix run github:astro/microvm.nix/alioth#alioth-example
```


pkgs/alioth.nix
```diff
@@ -0,0 +1,29 @@
+# TODO: upstream to nixpkgs once it no longer requires rust nightly
```
SuperSandro2000, line 1: We can probably hack around this by using the RUSTC_BOOTSTRAP env

PR #258, opened 2024-07-21
virtiofsd: use type notify

SuperSandro2000:


nixos-modules/host/default.nix
```diff
@@ -294,6 +339,10 @@ in
             " >> /tmp/supervisord.conf
           done
 
+          echo "[group:virtiofsd]
```
PatrickDaG, line 342: `supervisord` does not accept trailing commas.

nixos-modules/host/default.nix
```diff
@@ -294,6 +339,10 @@ in
...
+          programs=$group_programs
```
PatrickDaG: supervisord does not accept trailing commas in the program list.

nixos-modules/host/default.nix
```diff
@@ -267,18 +267,63 @@ in
...
           RestartSec = "5s";
           SyslogIdentifier = "microvm-virtiofsd@%i";
+          Type = "notify";
           WorkingDirectory = "${stateDir}/%i";
         };
         path = with pkgs; [ virtiofsd ];
         script = ''
           echo "[supervisord]
+          [eventlistener:virtiofsd]
+          command=${pkgs.writers.writePython3 "supervisord-event-handler" {
```
astro: Can the Python script be moved into a separate file?

astro (changes requested):

PR #267, opened 2024-08-11
Fix quoting of virtiofs sources in supervisord config

c0deaddict:
The quotes around `$SOURCE` need to be escaped, currently it breaks virtiofs sources which contain a space in the path.


astro:
Thank you!

Issue #284, opened 2024-10-13
macvtap-interfaces isn't written when `macvtap.mode == null`


inhumantsar:
```nix
microvm = {
    host.enable = true;
    vms = {
      "myvm" = {
        autostart = true;
        ...
        config = {
          microvm = {
            interfaces = [
                {
                  type = "macvtap";
                  id = "myvm";
                  mac = "somemac";
                  macvtap.link = "eno1";
                }
            ];
            ...
```

with a config like the one above, microvm will seem to apply fine but `/var/lib/microvms/myvm/current/share/microvm/macvtap-interfaces` won't be written and so the VM will fail to start.

then i noticed [this](https://github.com/astro/microvm.nix/blob/5e77efdc9ae9c8486814ebc1282d36e4cf780f30/lib/runner.nix#L87) and realized what was happening. after adding `macvtap.mode = "vepa"` and rebuilding, the `macvtap-interfaces` file was written successfully and the interface was created.

the default mode for macvtap interfaces in systemd is `vepa` and the default mode in microvm is `null`. the option description doesn't indicate that this option is required when `type == "macvtap"`, so i assumed that meant it would use the system default (ie: `vepa`).

seems like microvm should either default to `macvtap.mode = vepa` when `type == "macvtap"` or throw an error during evaluation if  `macvtap.mode == null && type == "macvtap"`.

PR #285, opened 2024-10-13
lib/runner: add macvtap config checks

astro:
@inhumantsar Does this solve #284?


inhumantsar:
I've stepped away from the computer for the evening but that looks like it would do the trick! I'll give it a try tomorrow or Monday and let you know.

Thanks!

astro:
Please test, I don't use macvtap myself but tap and sr-iov.

astro:
This is no longer required as the macvtap options are no longer `types.nullOr`. Their missing default does no longer cause an error when they're not used in case of `type != "macvtap"`. Was there an improvement in the NixOS module system?

Issue #273, opened 2024-09-10
Block device mount in guest (ZFS zvol)


pierregillet:
I would like to access a ZFS zvol (block device) from the VM but can't figure it out for the life of me.
I saw references to block devices mounting in the documentation, and this comment that was posted:

https://astro.github.io/microvm.nix/faq.html#how-do-i-let-the-microvm-user-access-block-devices

    As there are multiple solutions to allowing block device access (user extraGroup or udev rules), we cannot check disk permissions at build time. At runtime the VMM will error anyway.

    I've added some documentation. The next release will be breaking anyway.

_Originally posted by @astro in https://github.com/astro/microvm.nix/issues/222#issuecomment-2027637863_

To mount it, I have tried :

1. Adding microvm user to `disk` group

          users.users.microvm.extraGroups = [ "disk" ];
2. Then:
   * mounting my zvol as a microvm volume (does not seem correct, as volumes are only for FS mount, right ?)
      ```
      {
        mountPoint = "/mnt/myvol";
        image = "/dev/zvol/mypool/myvol";
      }
      ```
   * as a last solution, passing extra args to qemu 
      ```
     microvm = {
         qemu.extraArgs = [
         # "-drive file=/dev/zvol/mypool/myzvol,format=raw,if=virtio"  # first try
         # "-drive file=/dev/zvol/mypool/myzvol,format=raw"  # second try
         # "-blockdev node-name=q1,driver=raw,file.driver=host_device,file.filename=/dev/zvol/mypool/myzvol -device virtio-blk,drive=q1"   # third try
         # ];
     };
      ```

Would you have any pointers on how to proceed ? 

(disclaimer: I am rather new to nixos, and qemu, may not have seen the obvious)

astro:
The extra group and the former config look right. What's the error message you are getting?

pierregillet:
When I use the volume as described, without the `size` attribute, I get the following error:

    error: The option `microvm.volumes."[definition 1-entry 3]".size' is used but not defined.

When I set a dummy volume size (`10240` in this case), the VM is built and starts as expected, and I don't see any error. 
However the volume does not seem to be mounted.

I don't see it mounted  in the output of `lsblk` nor `mount` (in the VM)

astro:
That's weird because `microvm.volumes` end up in `fileSystems`: https://github.com/astro/microvm.nix/blob/main/nixos-modules/microvm/mounts.nix#L96

pierregillet:
[...]

PR #287, opened 2024-10-13
support unmounted volumes

pierregillet:
Fixes/implements https://github.com/astro/microvm.nix/issues/273

Allow for a block device from the host to be accessible in the guest without mounting it as a filesystem in the guest.

### DIFF
--- a/nixos-modules/microvm/mounts.nix
+++ b/nixos-modules/microvm/mounts.nix
@@ -110,7 +110,7 @@ lib.mkIf config.microvm.guest.enable {
       }) {} (withDriveLetters config.microvm)
   ) (
     # 9p/virtiofs Shares
-    builtins.foldl' (result: { mountPoint, tag, proto, source, ... }: result // {
+    builtins.foldl' (result: { mountPoint, tag, proto, source, ... }: result // lib.optionalAttrs (mountPoint != null) {
       "${mountPoint}" = {
         device = tag;
         fsType = proto;


nixos-modules/microvm/mounts.nix
```diff
@@ -110,7 +110,7 @@ lib.mkIf config.microvm.guest.enable {
       }) {} (withDriveLetters config.microvm)
   ) (
     # 9p/virtiofs Shares
```
astro, line 112: shares != volumes

Is there a use-case for not mounted 9p/virtiofs?

astro (changes requested):

nixos-modules/microvm/mounts.nix
pierregillet: I am unsure of the semantics between shares & volumes here ?

This PR allows me to mount my ZFS zpool inside the guest, without expecting a filesystem. 
I set this config in the guest machine configuration:
```
microvm = {
  volumes = [
    {
      mountPoint = null;
      image = "/dev/disk/by-label/my-zpool";
      size = 1024;
    }
  ];
};
```

My use case is that I have a service expecting root access to a whole ZFS zpool in the VM.

astro: - volumes: Block devices
- shares: Directories

A ZVOL should be used as a volume. `microvm.volumes` already allows for `mountPoint = null`.

astro: Does that distinction clarify your issue?

pierregillet: It seems it was indeed supported all along and I wrote an erroneous configuration the first time around, giving me wrong assumptions.
PEBCAK issue.

Thank you for taking the time, appreciated

PR #288, opened 2024-10-19
Fix tapScript doc rendering

SuperSandro2000:
see https://search.nixos.c3d2.de/?scope=6&option=microvm.host.tapScript

### DIFF
--- a/nixos-modules/host/options.nix
+++ b/nixos-modules/host/options.nix
@@ -40,6 +40,7 @@
       default = ''
         ${pkgs.iproute2}/bin/ip link set "$id" up
       '';
+      defaultText = "''${pkgs.iproute2}/bin/ip link set \"$id\" up";
     };
 
     vms = mkOption {


nixos-modules/host/options.nix
```diff
@@ -37,9 +37,8 @@
         ''${pkgs.iproute2}/bin/ip link set "$id" master br0 up
       '';
       type = types.lines;
-      default = ''
-        ${pkgs.iproute2}/bin/ip link set "$id" up
-      '';
+      default = "${pkgs.iproute2}/bin/ip link set "$id" up";
+      defaulText = "''${pkgs.iproute2}/bin/ip link set "$id" up";
```
```suggestion
      defaultText = "''${pkgs.iproute2}/bin/ip link set "$id" up";
```

astro (changes requested):
Why did this PR have draft status?

nixos-modules/host/options.nix
```diff
@@ -37,9 +37,8 @@
         ''${pkgs.iproute2}/bin/ip link set "$id" master br0 up
       '';
       type = types.lines;
-      default = ''
-        ${pkgs.iproute2}/bin/ip link set "$id" up
-      '';
+      default = "${pkgs.iproute2}/bin/ip link set "$id" up";
+      defaultText = "''${pkgs.iproute2}/bin/ip link set "$id" up";
```
```suggestion
      defaultText = "''${pkgs.iproute2}/bin/ip link set \"$id\" up";
```

SuperSandro2000:
> Why did this PR have draft status?

@codingmarcel is supposed to look, if it is now what we expect

nixos-modules/host/options.nix
```diff
@@ -37,9 +37,8 @@
         ''${pkgs.iproute2}/bin/ip link set "$id" master br0 up
       '';
       type = types.lines;
-      default = ''
-        ${pkgs.iproute2}/bin/ip link set "$id" up
-      '';
+      default = "${pkgs.iproute2}/bin/ip link set "$id" up";
```
MarcelCoding:
```suggestion
      default = "${pkgs.iproute2}/bin/ip link set \"$id\" up";
```

nixos-modules/host/options.nix
```diff
@@ -37,9 +37,8 @@
         ''${pkgs.iproute2}/bin/ip link set "$id" master br0 up
       '';
       type = types.lines;
-      default = ''
-        ${pkgs.iproute2}/bin/ip link set "$id" up
-      '';
+      default = "${pkgs.iproute2}/bin/ip link set \"$id\" up";
```
astro: Can we keep the `default = ''` with a newline, now that we have the `defaultText` too?

astro (changes requested):

PR #289, opened 2024-10-20
doc: fix literalExpression in examples

MarcelCoding:
Before:
image

After:
image

### DIFF
--- a/nixos-modules/host/options.nix
+++ b/nixos-modules/host/options.nix
@@ -32,15 +32,17 @@
         at all, just set
         `systemd.services."microvm-tap-interfaces@%i.service".enable = false`
       '';
-      example = ''
+      example = lib.literalExpression ''
         # Attach tap interface to bridge br0, and bring it up
-        ''${pkgs.iproute2}/bin/ip link set "$id" master br0 up
+        "${pkgs.iproute2}/bin/ip link set \"$id\" master br0 up"
       '';
       type = types.lines;
       default = ''
         ${pkgs.iproute2}/bin/ip link set "$id" up
       '';
-      defaultText = "''${pkgs.iproute2}/bin/ip link set \"$id\" up";
+      defaultText = ''
+        ${pkgs.iproute2}/bin/ip link set "$id" up
+      '';
     };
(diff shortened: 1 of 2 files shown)


nixos-modules/host/options.nix
```diff
@@ -32,14 +32,13 @@
...
+      example = lib.literalExpression ''
         # Attach tap interface to bridge br0, and bring it up
-        ''${pkgs.iproute2}/bin/ip link set "$id" master br0 up
+        "${pkgs.iproute2}/bin/ip link set \"$id\" master br0 up"
       '';
       type = types.lines;
-      default = ''
-        ${pkgs.iproute2}/bin/ip link set "$id" up
-      '';
+      default = "${pkgs.iproute2}/bin/ip link set \"$id\" up";
```
astro: Can we keep the `default = ''` with a newline, now that we have the `defaultText` too?

nixos-modules/microvm/options.nix
```diff
@@ -148,8 +148,7 @@ in
           };
         });
       default = [];
-      example = lib.literalExpression
-        ''
+      example = lib.literalExpression /* nix */ ''
```
astro, line 151: I like removing the linebreak here but what about the `/* nix */`? Not gonna enforce a `nixfmt` in this project! :laughing:

astro (changes requested):

nixos-modules/microvm/options.nix
MarcelCoding: It's tree sitter magic that allows tree sitter to inject a language in the string.

nixos-modules/host/options.nix
```diff
@@ -32,14 +32,13 @@
...
+      defaultText = "${pkgs.iproute2}/bin/ip link set \"$id\" up";
```
```suggestion
      default = ''
        ${pkgs.iproute2}/bin/ip link set "$id" up
      '';
      defaultText = ''
        ${pkgs.iproute2}/bin/ip link set "$id" up
      '';
```

astro:
Cheers!

PR #291, opened 2024-10-26
nixos-modules: move virtiofsd from host to microvm

astro:
@SuperSandro2000 Can you review pls? Note the simplified `virtiofsdCount` logic.

### DIFF
--- a/nixos-modules/host/default.nix
+++ b/nixos-modules/host/default.nix
@@ -161,41 +161,14 @@ in
         description = "Setup MicroVM '%i' devices for passthrough";
         before = [ "microvm@%i.service" ];
         partOf = [ "microvm@%i.service" ];
-        unitConfig.ConditionPathExists = "${stateDir}/%i/current/share/microvm/pci-devices";
+        unitConfig.ConditionPathExists = "${stateDir}/%i/current/bin/pci-setup";
         restartIfChanged = false;
         serviceConfig = {
           Type = "oneshot";
           RemainAfterExit = true;
           SyslogIdentifier = "microvm-pci-devices@%i";
+          ExecStart = "${stateDir}/%i/current/bin/pci-setup";
         };
-        # `ExecStart`
-        scriptArgs = "%i";
-        script = ''
-          cd ${stateDir}/$1
-
-          ${pkgs.kmod}/bin/modprobe vfio-pci
-
-          for path in $(cat current/share/microvm/pci-devices); do
-            pushd /sys/bus/pci/devices/$path
-            if [ -e driver ]; then
-              echo $path > driver/unbind
-            fi
-            echo vfio-pci > driver_override
-            echo $path > /sys/bus/pci/drivers_probe
-
-            # In order to access the vfio dev the permissions must be set
-            # for the user/group running the VMM later.
-            #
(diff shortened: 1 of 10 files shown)


nixos-modules/host/default.nix
```diff
@@ -250,68 +250,37 @@ in
...
-          "--inode-file-handles=${config.microvm.virtiofsd.inodeFileHandles}"
-          } \
-          ${lib.concatStringsSep " " config.microvm.virtiofsd.extraArgs}
-          " >> /tmp/supervisord.conf
-          done
+      "microvm-virtiofsd@" =
+        let
+          runFromBootedOrCurrent = name: pkgs.writeShellScript "virtiofsd-${name}" ''
+            if [ -e booted ]; then
+              exec booted/bin/${name}
```
SuperSandro2000: Can we easily do absolute paths here? That would make testing things by hand a lot easier.

nixos-modules/microvm/virtiofsd/default.nix
```diff
@@ -0,0 +1,72 @@
...
+
+in
+{
+  microvm.virtiofsdScripts = lib.mkIf requiresVirtiofsd {
+    run =
+      let
+        supervisordConfig = pkgs.writeText "${config.networking.hostName}-virtiofsd-supervisord.conf" ''
+          [supervisord]
+          nodaemon=true
+          #user=root
```
SuperSandro2000: I think it required that explicitly but I would need to check the docs

nixos-modules/microvm/virtiofsd/default.nix
```diff
@@ -0,0 +1,72 @@
...
+
+  # TODO: don't hardcode
+  group = "kvm";
+
+in
+{
+  microvm.binScripts = lib.mkIf requiresVirtiofsd {
+    virtiofsd-run =
+      let
+        supervisordConfig = pkgs.writeText "${config.networking.hostName}-virtiofsd-supervisord.conf" ''
```
MarcelCoding: How about using the ini renderer?

MarcelCoding (approved):

MarcelCoding:
Lgtm, did not test

MarcelCoding:
How should the migration be done. All vms need to be stopped, mciro m in the host needs to be updated, all vms need to be rebuild and started

astro:
> How should the migration be done. All vms need to be stopped, mciro m in the host needs to be updated, all vms need to be rebuild and started

No urgent requirement to stop VMs but they must be updated before they can be started again. Hence the warning that is output on host activation. My hope is that VMs are updated more frequently than hosts.

MarcelCoding:
How about a check if the files exists and if yes output that a rebuild is required.

MarcelCoding:
Or create a new major release.

nixos-modules/microvm/virtiofsd/default.nix
astro: It's not required. That's great because that means it won't fail when the user runs it from a package instead of thru the systemd unit.

astro:
> How about a check if the scripts exists and if yes output that a rebuild is required.

I am not sure I understand, the activationScript snippet does check for the scripts' existence.

> Or create a new major release.

I've prepared the CHANGELOG.md entries for that but let's wait a bit for any regressions.

PR #292, opened 2024-10-31
Add support for remote sudo

wokalski:


astro:
The parameter is not exactly position-independent but useful nevertheless!

PR #295, opened 2024-11-05
Add direct and serial options to disk devices

wokalski:
I only added support for cloud-hypervisor. The options are also supported in qemu (and possibly others) but i am just using cloud hypervisor.

serial is not documented yet, but it's been supported by cloud hypervisor for a year.

One thing to watch out for is the length of serial, it has to be 20 characters (or less, maybe).

Direct does not improve I/O performance per se, it mostly reduces overhead and makes VMs less noisy (O_DIRECT).

Maybe this PR should be closed wihout merging and we should have a manual way (i.e. a blanket "extraOpts" option) to specify hypervisor specific options because the doesn't pattern doesnt seem sustainable.

I am not using linux and I didn't want to spend time on setting up the formatter for nix (I am usually using nixfmt in my flake). So if you would like to fix formatting please rebase and format.

### DIFF
--- a/lib/runners/cloud-hypervisor.nix
+++ b/lib/runners/cloud-hypervisor.nix
@@ -26,7 +26,11 @@ let
   useVirtiofs = builtins.any ({ proto, ... }: proto == "virtiofs") shares;
 
   # Transform attrs to parameters in form of `key1=value1,key2=value2,[...]`
-  opsMapped = ops: lib.concatStringsSep "," (lib.mapAttrsToList (k: v: "${k}=${v}") ops);
+  opsMapped = ops: lib.concatStringsSep "," (
+    lib.mapAttrsToList
+      (k: v: "${k}=${v}")
+      (lib.attrsets.filterAttrs (_: v: v != null) ops)
+  );
 
   # Attrs representing CHV mem options
   memOps = opsMapped ({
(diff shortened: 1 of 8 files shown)


lib/runners/cloud-hypervisor.nix
```diff
@@ -26,7 +26,11 @@ let
   useVirtiofs = builtins.any ({ proto, ... }: proto == "virtiofs") shares;
 
   # Transform attrs to parameters in form of `key1=value1,key2=value2,[...]`
-  opsMapped = ops: lib.concatStringsSep "," (lib.mapAttrsToList (k: v: "${k}=${v}") ops);
+  opsMapped = ops: lib.concatStringsSep "," (
+    lib.mapAttrsToList
+      (k: v: "${k}=${v}")
+      (lib.attrsets.filterAttrs (_: v: v != null) ops)
```
astro: `null` could also mean a key without value which isn't obvious. I actually prefer an immediate `} // lib.optionalAttrs (v != null) { ${k} = v;` in the caller. What do you think?

nixos-modules/microvm/options.nix
```diff
@@ -185,6 +185,18 @@ in
...
+            default = null;
+            description = "User-configured serial number for the disk
+            (Currently only respected if using cloud-hypervisor)";
+          };
+          direct = mkOption {
+            type = nullOr bool;
+            default = null;
+            description = "Whether to set O_DIRECT on the disk.
+            (Currently only respected if using cloud-hypervisor)";
+          };
```
astro, line 199: Makes sense and I'm going to merge it, but only after someone has checked and maybe implemented support for the other hypervisors. :-)

astro (approved):

lib/runners/cloud-hypervisor.nix
wokalski: Sure. Will fix it.

astro:
@wokalski Could you please review?

PR #303, opened 2024-12-03
fix macvtaps

PatrickDaG:
#291 introduced two bugs related to macvtaps:
- multiple definition of `tapMultiQueue` in the passthru attributes.
I'm guessing the on I removed shouldn't be there as it's just the other one without
the default case.
- wrong type of `binScripts.macvtap-down`


RooSoft:
This breaks my setup, had to revert to a prior commit... thanks guys for working this through.

I guess I would be better off not refering to the bleeding edge...

astro:
macvtap is a feature I do not test myself. Thank you very much for doing this for me!

I create git tags from time to time after new features have stabilized.

RooSoft:
Working now with the bleeding edge, thanks!

PR #310, opened 2024-12-12
qemu: support notify sockets

vikanezrimaya:
Copied from existing cloud-hypervisor implementation. Tested by launching a MicroVM on a system that expects MicroVMs to use notify sockets, and additionally using plain socat in a second terminal.


astro:
Awesome! Thank you.

PR #313, opened 2024-12-13
WIP: qemu: randomly choose vsock port on the host side

vikanezrimaya:
Prevents conflicts between MicroVMs.

Somehow doesn't work first try; VM might timeout a few times before it starts working. The problem is most likely between socat and systemd. More debug logging is needed. (Adding it here would be trivial, actually.)

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -147,6 +147,7 @@ let
     else "";
 
   supportsNotifySocket = vsock.cid != null;
+  vsockHostPortOffset = 600000;
 
 in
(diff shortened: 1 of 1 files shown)


astro:
How about creating the port number from bash `$RANDOM` instead of reading it back from socat's log?

astro:
Alternatively, `microvm.vsock.cid` should be unique per VM. Let's just add that to an offset.

astro:
Are you going to work on this on the next few days, or may I roll back the qemu change?

PR #323, opened 2025-01-14
Fixed the missing reference to kernelAtLeast for erofs

Michael-C-Buckley:
Fixed a bug where `nixos-modules.options.nix` did not have a reference for `kernelAtLeast`, which was instead present in `nixos-modules.store-disk.nix`, where it wasn't needed.  It had only seemed to affect `erofs` volumes, where the reference was needed.


astro:
Thanks but duplicate of #322.

PR #324, opened 2025-01-23
Add deflate-on-oom and free-page-reporting for qemu

slakkala:
Enable automatic balloon deflation on QEMU. cloud-hypervisor already enables it. Also enable free-page-reporting, and give the balloon device a name for easier querying.

Additionally make OOM deflation configurable.


nixos-modules/microvm/options.nix
```diff
@@ -466,6 +466,14 @@ in
       '';
     };
 
+    qemu.oomDeflate = mkOption {
```
astro: [It looks like this feature is supported on more VMMs than just qemu.](https://github.com/firecracker-microvm/firecracker/blob/main/docs/ballooning.md) Let's make it generic:

```suggestion
    oomDeflate = mkOption {
```

Also, it has the same name `deflate_on_oom` so I wonder if `deflateOnOOM` is a more meaningful name for the option. YMMV.

astro (changes requested):
Thank you!

nixos-modules/microvm/options.nix
slakkala: Changed to deflateOnOOM and dropped the `qemu`

astro (approved):

PR #325, opened 2025-01-24
qemu: add deviceExtraArgs option for devices

vunnyso:
This option could be useful in case we may need to pass any extra arguments to pci device.
For example:
`-device vfio-pci,host=x,multifunction=on,id=hostId,x-igd-opregion=on`

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -290,9 +290,12 @@ lib.warnIf (mem == 2048) ''
       "-device" "qemu-xhci"
     ]
     ++
-    builtins.concatMap ({ bus, path, ... }: {
+    builtins.concatMap ({ bus, path, qemu,... }: {
       pci = [
-        "-device" "vfio-pci,host=${path},multifunction=on"
+        "-device" "vfio-pci,host=${path},multifunction=on${
+          # Allow to pass additional arguments to pci device
+          lib.optionalString (qemu.deviceExtraArgs != null) ",${qemu.deviceExtraArgs}"
+        }"
       ];
       usb = [
         "-device" "usb-host,${path}"
--- a/nixos-modules/microvm/options.nix
+++ b/nixos-modules/microvm/options.nix
@@ -352,6 +352,7 @@ in
         } {
           bus = "pci";
           path = "0000:01:01.0";
+          deviceExtraArgs = "id=hostId,x-igd-opregion=on";
         } {
           # QEMU only
           bus = "usb";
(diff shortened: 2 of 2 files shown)


vunnyso:
@astro please help to review.

vunnyso:
Rebased.

nixos-modules/microvm/options.nix
```diff
@@ -372,6 +373,13 @@ in
               Identification of the device on its bus
             '';
           };
+          deviceExtraArgs = mkOption {
```
astro: I am a bit nitty about the name: the option is already inside `devices` and it's qemu-specific. How do you like `qemu.extraArgs`?

lib/runners/qemu.nix
```diff
@@ -290,9 +290,12 @@ lib.warnIf (mem == 2048) ''
       "-device" "qemu-xhci"
     ]
     ++
-    builtins.concatMap ({ bus, path, ... }: {
+    builtins.concatMap ({ bus, path, deviceExtraArgs,... }: {
       pci = [
-        "-device" "vfio-pci,host=${path},multifunction=on"
+        "-device" "vfio-pci,host=${path},multifunction=on${
+          # Allow to pass additional arguments to pci device
+          lib.optionalString (requirePci && deviceExtraArgs != null) ",${deviceExtraArgs}"
```
astro: `requirePci` is already `true` when there are pci `devices`.

astro (changes requested):

nixos-modules/microvm/options.nix
vunnyso: I think `qemu.extraArgs` allow to pass extra arguments to `qemu`
**Example:**
` -device 'tpm-tis,tpmdev=tpm' `

And what I am trying to add is allowing to pass extra args to `pci` device itself

**Example:** 
Say for graphics `pci` device, if I want to pass extra arguments to particular device then I think below args comes by default:
` -device 'vfio-pci,host=0000:00:02.0,multifunction=on' `

In case say I want to pass extra `pci` arguments to graphics device
**Example:**
` -device 'vfio-pci,host=0000:00:02.0,multifunction=on,id=hostId,x-igd-opregion=on' `

We can do if deviceExtraArgs option is allowed.

lib/runners/qemu.nix
vunnyso: Cool, I have removed extra check.

nixos-modules/microvm/options.nix
astro: I was proposing renaming `deviceExtraArgs` into `qemu.extraArgs` but obviously that would be very confusing... :-) `qemu.deviceExtraArgs`?

vunnyso: Addressed.

astro:
Thanks

Issue #326, opened 2025-01-29
evaluation of derivation fails if BalloonMemory is not 0


71rd:
Evaluation fails if any vm has a value bigger than 0 for balloon memory, since merging PR #324.

In that PR the deflate_on_oom option was moved into a separate optional set from the other balloon options, but when setting balloon options the other set of options is converted into a list before trying to merge it with the new set, instead of first merging the sets before converting to the final list of used options.

PR #327, opened 2025-01-29
fix: evaluation wont fail when BalloonMemory is enabled

71rd:
When BalloonMemory was not 0 evaluation of derivations failed after PR #324 made deflate-on-oom optional.
The new optional set containing the argument was only merged with the other arguments, after those were converted into a list leading to a type mismatch.

fixes: #326


astro:
Thank you very much!

PR #328, opened 2025-01-31
fix: replace deprecated hardware.opengl.enable

timon-schelling:
replace deprecated hardware.opengl.enable with hardware.graphics.enable

removes warning: The option `hardware.opengl.enable' defined in `/nix/store/...-source/flake.nix' has been renamed to `hardware.graphics.enable'.


astro:
Thank you!

PR #331, opened 2025-02-05
cloud-hypervisor: use absolute paths for sockets

SuperSandro2000:
so that scripts can be executed from anywhere


astro:
Being able to use relative paths for sockets, volumes, and shares is an actual feature.

astro:
As users we are free to set `microvm.socket` to an absolute path.

PR #332, opened 2025-02-06
cloud-hypervisor: fix eval when deflateOnOOM is off

SuperSandro2000:


astro:
D'oh, I should take more time for manual testing submissions... thank you!

PR #336, opened 2025-02-18
cloud-hypervisor: add platformOEMStrings and `--platform` merging

Ramblurr:
The cloud-hypervisor command line interface unfortunately doesn't support
multiple instances of the same arg with a different value, so we have to resort
to these extra module options rather than using extraArgs.

To make matters even worse, the `--platform` argument (of which there can be
only one), is overloaded with different types of sub-args that also need to be
provided multiple times.

This commit allows the operator to add oem strings (for example to pass systemd
credentials), as well as raw platform options as needed.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -107,7 +107,7 @@ rec {
       => { values = ["linux"]; args = ["-vnc" ":0" "-usb"]; }
 
       # Extract multiple occurrences:
-      extractOptValues "-platform" ["-a" "a" "-b" "b" "-c" "c" "-b" "b2"]
+      extractOptValues "-b" ["-a" "a" "-b" "b" "-c" "c" "-b" "b2"]
       => { values = ["b" "b2"]; args = ["-b" "b" "-c" "c"]; }
 
       # Extract with multiple flag aliases:
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -108,7 +108,7 @@ rec {
 
       # Extract multiple occurrences:
       extractOptValues "-b" ["-a" "a" "-b" "b" "-c" "c" "-b" "b2"]
-      => { values = ["b" "b2"]; args = ["-b" "b" "-c" "c"]; }
+      => { values = ["b" "b2"]; args = ["-a" "a" "-c" "c"]; }
 
       # Extract with multiple flag aliases:
       extractOptValues ["-p" "-platform"] ["-p" "short" "-vnc" ":0" "-platform" "long" "-usb"]
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -78,4 +78,59 @@ rec {
     import ./macvtap.nix {
       inherit microvmConfig hypervisorConfig lib;
     };
+
+  /*
+    extractOptValues - Extract and remove all occurrences of a command-line option and its values from a list of arguments.
+
+    Description:
+      This function searches for a specified option flag in a list of command-line arguments,
+      extracts ALL associated values, and returns both the values and a filtered list with
+      all occurrences of the option flag and its values removed. The order of all other
+      arguments is preserved. Uses tail recursion to process the argument list.
+
+    Parameters:
+      optFlag :: String | [String] - The option flag(s) to search for. Can be:
+                                     - A single string (e.g., "-platform")
+                                     - A list of strings (e.g., ["-p" "-platform"])
+                                     All matching flags and their values are extracted
+      extraArgs :: [String] - A list of command-line arguments
+
+    Returns:
+      {
+        values :: [String] - List of all values associated with matching flags (empty list if none found)
+        args :: [String] - The input list with all matched flags and their values removed
+      }
+
+    Examples:
+      # Extract single occurrence:
+      extractOptValues "-platform" ["-vnc" ":0" "-platform" "linux" "-usb"]
+      => { values = ["linux"]; args = ["-vnc" ":0" "-usb"]; }
+
+      # Extract multiple occurrences:
+      extractOptValues "-platform" ["-a" "a" "-b" "b" "-c" "c" "-b" "b2"]
+      => { values = ["b" "b2"]; args = ["-b" "b" "-c" "c"]; }
+
+      # Extract with multiple flag aliases:
+      extractOptValues ["-p" "-platform"] ["-p" "short" "-vnc" ":0" "-platform" "long" "-usb"]
+      => { values = ["short" "long"]; args = ["-vnc" ":0" "-usb"]; }
+
+      # Degenerate case with no matches:
+      extractOptValues ["-p" "-platform"] ["-vnc" ":0" "-usb"]
+      => { values = []; args = ["-vnc" ":0" "-usb"]; }
+  */
+  extractOptValues = optFlag: extraArgs:
+    let
+      flags = if builtins.isList optFlag then optFlag else [optFlag];
+
+      processArgs = args: values: acc:
+        if args == [] then
+          { values = values; args = acc; }
+        else if (builtins.elem (builtins.head args) flags) && (builtins.length args) > 1 then
+          # Found one of the option flags, skip it and its value
+          processArgs (builtins.tail (builtins.tail args)) (values ++ [(builtins.elemAt args 1)]) acc
+        else
+          # Not the option we're looking for, keep this element
+          processArgs (builtins.tail args) values (acc ++ [(builtins.head args)]);
+    in
+      processArgs extraArgs [] [];
 }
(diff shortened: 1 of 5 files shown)


nixos-modules/microvm/options.nix
```diff
@@ -483,6 +483,30 @@ in
       '';
     };
 
+    cloud-hypervisor.platformOEMStrings = mkOption {
```
astro, line 518: Instead of creating another option, you could just use `map (lib.removePrefix "oem_strings=") (builtins.filter (lib.hasPrefix "oem_strings=") config.cloud-hypervisor.extraPlatformOpts))` to get these.

nixos-modules/microvm/options.nix
```diff
@@ -483,6 +483,30 @@ in
...
+      description = ''
+      Extra arguments to pass to cloud-hypervisor's --platform oem_strings= argument.
+
+      All the oem strings will be concatenated with a comma (,) and wrapped in oem_string=[].
+      '';
+      example = literalExpression /* nix */ ''
+      [ "io.systemd.credential:APIKEY=supersecret" ]
+      '';
+    };
+    cloud-hypervisor.extraPlatformOpts = mkOption {
```
astro: ~I like cloud-hypervisor myself, but one of the main features of this Flake is choice with a single change to `config.microvm.hypervisor`. At least qemu has similar parameters but I really would like to see this researched for all seven supported VMMs.~ Oh, `--platform` *is* chv-specific!

I wonder if this too should be filtered from `config.cloud-hypervisor.extraArgs`, thereby removing that limitation for us?

nixos-modules/microvm/options.nix
Ramblurr: That wouldn't quite parse it correctly because there is the square bracket list syntax too. And we have to combine it with the vsock OEM string. 

But we could parse them out and then rebuild the oem_strings arg, should I so that?

It's a shame that chv has such a clunky interface.

astro: You're the one with the new use-case here :-)

If you agree that it is worthwhile working around chv's clunky interface, so please go ahead!

If you prefer separate options, I'm going to accept them.

nixos-modules/microvm/options.nix
Ramblurr: The reason why I added the `extraPlatformOpts` option is because you can only specify `--platform` one time. But `--platform` has a whole world of options unto itself that we (via the vsock) or a user (me with my systemd credential wish)  would like to supply.

```
--platform <platform>      num_pci_segments=<num_pci_segments>,iommu_segments=<list_of_segments>,iommu_address_width=<bits>,serial_number=<dmi_device_serial_number>,uuid=<dmi_device_uuid>,oem_strings=<list_of_strings>
```

Your suggestion of filtering out the `--platform` from `config.cloud-hypervisor.extraArgs` is a good one. I will implement that and eliminate the need for a new `cloud-hypervisor.extraPlatformOpts`

nixos-modules/microvm/options.nix
Ramblurr: The oem_strings is tricky to handle, and I don't want the user to have to do the string manipulation themselves, and I also don't want to have to parse the array syntax just so we can put it back together with the vsock that we add. 

So I think it is best that a new option `microvm.cloud-hypervisor.platformOEMStrings` exists. But, as per my other comment, we can get rid of the `cloud-hypervisor.extraPlatformOpts`

Ramblurr:
Hi again @astro, I want to pick this little project back up.. so I'm going to be reviving these PRs today :) Please see my comments to your comments!.

I've just pushed new commits that should address your concerns.

nixos-modules/microvm/options.nix
Ramblurr: IMO this is resolved.

nixos-modules/microvm/options.nix
Ramblurr: IMO this is resolved!

lib/default.nix
```diff
@@ -78,4 +78,59 @@ rec {
...
+        args :: [String] - The input list with all matched flags and their values removed
+      }
+
+    Examples:
+      # Extract single occurrence:
+      extractOptValues "-platform" ["-vnc" ":0" "-platform" "linux" "-usb"]
+      => { values = ["linux"]; args = ["-vnc" ":0" "-usb"]; }
+
+      # Extract multiple occurrences:
+      extractOptValues "-platform" ["-a" "a" "-b" "b" "-c" "c" "-b" "b2"]
```
SuperSandro2000:
```suggestion
      extractOptValues "-b" ["-a" "a" "-b" "b" "-c" "c" "-b" "b2"]
```

If I understand this correctly, the example should have ``-b`` instead.

lib/runners/cloud-hypervisor.nix
```diff
@@ -5,8 +5,9 @@
 
 let
   inherit (pkgs) lib;
+  inherit (import ../. { inherit (pkgs) lib; }) extractOptValues;
```
SuperSandro2000: I am not a fan of this, but this existed before in other files.

SuperSandro2000: I opened https://github.com/microvm-nix/microvm.nix/pull/383 to change that in existing files. Can we adopt that pattern here?

nixos-modules/microvm/options.nix
```diff
@@ -515,6 +515,23 @@ in
...
+        --platform option to cloud-hypervisor
+      '';
+      example = lib.literalExpression /* nix */ ''
+          [ "io.systemd.credential:APIKEY=supersecret" ]
+        '';
+    };
+
     cloud-hypervisor.extraArgs = mkOption {
       type = with types; listOf str;
       default = [];
```
SuperSandro2000, line 537: Should we not in the description to not write the oem_strings from above into it? Maybe even assert it?

nixos-modules/microvm/options.nix
```diff
@@ -515,6 +515,23 @@ in
...
+
+        All the oem strings will be concatenated with a comma (,) and wrapped in oem_string=[].
+
+        The resulting string will be combined with any --platform options in
+        `config.microvm.cloud-hypervisor.extraArgs` and passed as a single
+        --platform option to cloud-hypervisor
+      '';
+      example = lib.literalExpression /* nix */ ''
+          [ "io.systemd.credential:APIKEY=supersecret" ]
+        '';
```
```suggestion
      example = lib.literalExpression /* nix */ ''[ "io.systemd.credential:APIKEY=supersecret" ]'';
```

lib/runners/cloud-hypervisor.nix
```diff
@@ -93,6 +94,11 @@ let
 
   supportsNotifySocket = true;
 
+  oemStringValues = (lib.optionals supportsNotifySocket ["io.systemd.credential:vmm.notify_socket=vsock-stream:2:8888"]) ++ platformOEMStrings;
+  oemStringOptions = lib.optionals  (oemStringValues != [])  ["oem_strings=[${lib.concatStringsSep "," oemStringValues}]"];
```
```suggestion
  oemStringValues = platformOEMStrings ++ lib.optional supportsNotifySocket "io.systemd.credential:vmm.notify_socket=vsock-stream:2:8888";
  oemStringOptions = lib.optional (oemStringValues != []) "oem_strings=[${lib.concatStringsSep "," oemStringValues}]";
```

lib/runners/cloud-hypervisor.nix
```diff
@@ -93,6 +94,11 @@ let
...
+  extraArgsWithoutPlatform = (extractOptValues "--platform" extraArgs).args;
+  userPlatformOpts = (extractOptValues "--platform" extraArgs).values;
```
SuperSandro2000: I am not sure how compute intensive extractOptValues is but maybe it would be worth it, to cache it?

Ramblurr:
@SuperSandro2000 I've pushed a new commit that rebases onto `main` to pickup your latest PR, and also addresses your comments. Let me know if there is anything else.

Ramblurr:
fixed the merge conflict

lib/default.nix
```diff
@@ -78,4 +78,59 @@ rec {
...
+      }
+
+    Examples:
+      # Extract single occurrence:
+      extractOptValues "-platform" ["-vnc" ":0" "-platform" "linux" "-usb"]
+      => { values = ["linux"]; args = ["-vnc" ":0" "-usb"]; }
+
+      # Extract multiple occurrences:
+      extractOptValues "-b" ["-a" "a" "-b" "b" "-c" "c" "-b" "b2"]
+      => { values = ["b" "b2"]; args = ["-b" "b" "-c" "c"]; }
```
SuperSandro2000:
```suggestion
      => { values = ["b" "b2"]; args = ["-a" "a" "-c" "c"]; }
```

SuperSandro2000 (approved):

SuperSandro2000:
Let's try this 🎉 thanks

Issue #259, opened 2024-07-27
Cannot get custom smbios values to work on qemu runner


Bert-Proesmans:
Trying to pass data from hypervisor into the VM's, according to [systemd's credential instructions](https://systemd.io/CREDENTIALS/#inheritance-from-container-managers-hypervisors-kernel-command-line-or-the-uefi-boot-environment) and [qemu invocation instructions](https://www.qemu.org/docs/master/system/invocation.html#hxtool-4)  
Assuming qemu "i386 target" means a family that contains x86_64 hosts, this all should work out of the box.

I'm not seeing the defined values inside the VM.

Steps to reproduce;
1. Create a barebones guest config with xxd and dmidecode packages (and configuration to get interactive console access)
2. Add the additional qemu invocation arguments `[ '-smbios' 'type=11,value=io.systemd.credential:mycred=supersecret' ]`
3. Launch guest vm on hypervisor host through qemu
4. Connect to guest
5. TEST; execute in guest `sudo systemd-creds --system cat mycred`
6. TEST; execute in guest `ls -la /sys/firmware/dmi/entries/11*`
7. TEST; execute in guest `sudo dmidecode -t 11`

Expected;
- systemd-creds should return the value supersecret
- ls should display folder contents of entry 11 (aka filesystem mapping of type-11 oem information)
- dmidecode should print the line io.systemd.credential:mycred=supersecret

Actual;
In my case none of the above happens. There are no credentials, no sys-fs mapping, nothing printed by dmidecode.

I've been searching around but there is very little concrete information i can find about why this could happen. I'm not sure if it's guest linux kernel config related, qemu configuration related, or host related.  
I'm writing this hoping you'd have some ideas where to look further, this is not really intended as a bugreport. **It would help me a lot already if anyone can confirm this "just works" on their end.**

Aside; Looking at boot logs from other people there is an explicit line saying `SMBIOS X.X present.` This lacks from my guest boot logs, but also there are dmi entries (sys-fs) and dmidecode does print the line `SMBIOS 2.4 present.` This is probably a red herring.

<details>

<summary>nix inputs</summary>

```text
[bert-proesmans@development:~/nix]$ nix flake metadata
Resolved URL:  git+file:///home/bert-proesmans/nix
Locked URL:    git+file:///home/bert-proesmans/nix
Description:   Bert Proesmans's NixOS configuration
Path:          /nix/store/49sbykkhk6hys0mjr4fcdp3pq7r98fpx-source
[...]

Issue #52, opened 2022-07-06
stable ssh keys


Sohalt:
Since `/etc/ssh/` is on the tmpfs, the VM generates a new ssh key-pair every time is gets rebooted.
I tried to create a virtiofsd share for `/etc/ssh` to keep the keys on the host system, but that makes sshd fail (presumably because there is an issue with the symlinks for the config).
Any idea why this would fail? Symlinks should work across filesystems and the directory should be there from the moment the machine starts, no?

astro:
I have my MicroVM's whole `/etc` on virtiofs on ZFS for that precise reason: stable ssh host keys. I guess the conflict is with NixOS' `/etc/static`?

If you think your solution is better practise we should add it to the handbook.

yangm97:
I guess truly stable ssh keys would need to be generated beforehand and applied using some secret management tool i.e. https://github.com/Mic92/sops-nix

Sohalt:
sops-nix uses the ssh host keys to decrypt all other secrets, so it's a chicken/egg problem

yangm97:
> sops-nix uses the ssh host keys to decrypt all other secrets, so it's a chicken/egg problem

I was thinking about bootstrapping using sops-nix on the host to store the vm ssh key and then mount the file secret read-only.

c0deaddict:
I use https://github.com/nix-community/impermanence/ to "solve" this problem. It can be used not only for SSH keys but also for other persistent state. I give each MicroVM a `/persist` shared via virtiofs, and then bind mount the relevant dirs/files:

 ```    services.openssh = {
      hostKeys = [
        {
          path = "/persist/etc/ssh/ssh_host_ed25519_key";
          type = "ed25519";
        }
        {
          path = "/persist/etc/ssh/ssh_host_rsa_key";
          type = "rsa";
          bits = 4096;
        }
      ];
    };

    fileSystems."/persist".neededForBoot = mkForce true;

    environment.persistence."/persist" = {
      directories = [
        "/var/lib/systemd/coredump"
        "/var/lib/nixos" # contains user/group id map
        "/var/log"
      ];

      files = [
        "/etc/machine-id"
        "/root/.bash_history"
      ];
    };```

matthew-salerno:
I've managed to do this with [impermanence](https://github.com/nix-community/impermanence) and [agenix-rekey](https://github.com/oddlama/agenix-rekey). Disregarding boiler plate, it looks like this:

Host
```nix
config.age = {
[...]

PR #337, opened 2025-02-18
feat: Add `microvm.credentialFiles` for passing credentials to guests

Ramblurr:
This commit implements `microvm.credentialFiles` a mechanism for passing
credentials into guest vms from the host.

Currently only support for qemu is implemented as I want to test the waters to
see if you're interested in this feature, Astro.

In addition to qmeu cloud-hypervisor can be supported via smbios. But it depends
on [this feature being added](https://github.com/cloud-hypervisor/cloud-hypervisor/issues/6951#issuecomment-2664967013), and also #336 being merged to microvm.nix

cloud-hypervisor could be supported immediately, but then the secrets would be
visible in the ps output.

A cursory code search shows that the following additional hypervisors could be
supported:

- crosvm: via fw_cfg, or smbios
- alioth: via fw_cfg
- stratovirt: via fw_cfg (maybe smbios)

kvmtool and firecracker both seem like they cannot be supported.

Related:
- #259
- #52

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -39,7 +39,7 @@ let
   qemu = overrideQemu (if microvmConfig.cpu == null then
     pkgs.qemu_kvm else pkgs.buildPackages.qemu_full);
 
-  inherit (microvmConfig) hostName cpu vcpu mem balloonMem deflateOnOOM user interfaces shares socket forwardPorts devices vsock graphics storeOnDisk kernel initrdPath storeDisk;
+  inherit (microvmConfig) hostName cpu vcpu mem balloonMem deflateOnOOM user interfaces shares socket forwardPorts devices vsock graphics storeOnDisk kernel initrdPath storeDisk credentialFiles;
   inherit (microvmConfig.qemu) machine extraArgs serialConsole;
 
   inherit (import ../. { inherit (pkgs) lib; }) withDriveLetters;
@@ -146,6 +146,8 @@ let
     then "console=ttyAMA0"
     else "";
 
+  systemdCredentialStrings = lib.mapAttrsToList (name: path: "name=opt/io.systemd.credentials/${name},file=${path}" ) credentialFiles;
+  fwCfgOptions = systemdCredentialStrings;
 
 in
 lib.warnIf (mem == 2048) ''
@@ -172,6 +174,9 @@ lib.warnIf (mem == 2048) ''
 
       "-chardev" "stdio,id=stdio,signal=off"
       "-device" "virtio-rng-${devType}"
+    ] ++
+    lib.optionals (fwCfgOptions != [])  [
+      "-fw_cfg" (lib.concatStringsSep "," fwCfgOptions)
     ] ++
     lib.optionals serialConsole [
       "-serial" "chardev:stdio"
--- a/nixos-modules/microvm/options.nix
+++ b/nixos-modules/microvm/options.nix
@@ -602,6 +602,18 @@ in
       description = "Flags to pass to gensquashfs";
       default = [ "-c" "zstd" "-j" "$NIX_BUILD_CORES" ];
     };
+
+    credentialFiles = mkOption {
+      type = with types; attrsOf path;
+      description = ''
+        Key-value pairs of credential files that will be loaded into the vm using systemd's io.systemd.credential feature.
+      '';
+      example = literalExpression /* nix */ ''
+        {
+          SOPS_AGE_KEY = "/run/secrets/guest_microvm_age_key";
+        }
+      '';
+    };
   };
 
   config = lib.mkMerge [ {


Ramblurr:
> Are you going to try adding the other hypervisors?

Yes I will attempt to implement it in the runners for the other hypervisors, and for those that I can't I'll add an assert. I'll also update the docs.

I'd like to implement a test for this too. Could you point me to where in `checks/` I should add it? I thought somewhere in `default.nix` but there's a lot going on in there.

FYI this is how I am able to take advantage of the credential, by setting the SSH host key (which also lets me bootstrap my sops-nix in the guest).

```nix
  # this is part of the guest vms nixos config
  microvm.credentialFiles = {
    "SSH_HOST_KEY" = "/run/secrets/mymicrovm_sops_ssh_key"; # This file must exist on the host, since I use imperative vms, I have to make it manually.
  };
  systemd.services.sshd = {
    serviceConfig = {
      ImportCredential = "SSH_HOST_KEY";
    };
    preStart = ''
      mkdir -p /etc/ssh
      cat $CREDENTIALS_DIRECTORY/SSH_HOST_KEY > /etc/ssh/ssh_host_ed25519_key
      chmod 0600 /etc/ssh/ssh_host_ed25519_key
    '';
  };
```

I got ill, so it might be until next week before I'm able to return to these PRs.

astro:
The top of `checks/default.nix` contains all the combinations to permute for a "test matrix". You can add add a list containing configurations with and without the feature. The idea is to be able to test for any feature combination but it is not yet CI as of today. Related: https://github.com/nix-community/infra/pull/1598

Get well soon, and don't feel stressed from open tickets!

Sveske-Juice:
Is there a workaround currently (until this gets implemented) for passing sops-nix secrets through to the vm? I tried creating a shared directory by mounting `/run/secrets-for-users` but i get a permission error:
```
microvm@<vmname>: -device virtio-9p-pci,fsdev=fs1,mount_tag=secrets: cannot initialize fsdev 'fs1': failed to open '/run/secrets-for-users': Permission Denied
```
This is what I tried:
```nix
# ... other config
microvm.shares = [
  # nix read only store ...
  {
    source = "/run/secrets-for-users";
    mountPoint = "/run/secrets-for-users";
    tag = "secrets";
    proto = "9p";
  }
];
```

astro:
@Sveske-Juice We use sops-nix with per-VM keys.

Sveske-Juice:
> @Sveske-Juice We use sops-nix with per-VM keys.

So if I am understanding right. If this PR gets merged. We would be able to decrypt the secrets via the VM's key. So this feature would make it possible to bootstrap the VM with a valid SSH host key that can be used to decrypt the secrets?

Ramblurr:
> The top of `checks/default.nix` contains all the combinations to permute for a "test matrix". You can add add a list containing configurations with and without the feature

Ok I am well again and have some time available for FOSS contributions :) 

I'm trying to just get my test running ([see the commit](https://github.com/astro/microvm.nix/pull/337/commits/89e5bc16c4effb89a365a31878c5393808be9553) I just pushed), but when i run `nix flake check`, i get:

```
       … while calling the 'throw' builtin
         at /nix/store/vqhkmj60457j8nrw9vb94ln40rkic1p6-source/lib/modules.nix:859:9:
          858|         # handling.  If changed here, please change it there too.)
          859|         throw "The option `${showOption loc}' was accessed but has no value defined. Try setting the option.";
             |         ^
          860|
       error: The option `microvm.credentialFiles' was accessed but has no value defined. Try setting the option.
```

Any  tips?

lib/runners/qemu.nix
```diff
@@ -146,6 +146,8 @@ let
     then "console=ttyAMA0"
     else "";
 
+  systemdCredentialStrings = lib.mapAttrsToList (name: path: "name=opt/io.systemd.credentials/${name},file=${path}" ) credentialFiles;
+  fwCfgOptions = systemdCredentialStrings;
```
Bert-Proesmans, line 159: @Ramblurr 

Usage of the "credentialFiles" configuration comes down to this equivalent in another language `for name, value in credentialFiles; do {} end`, but credentialFiles holds NULL. More specifically, there are zero value assignments for this option in the whole configuration.

In other words, going forward everyone is forced to define option value `microvm.credentialFiles = {}` (at a minimum) to keep their configuration compiling. Hence the (unrelated) error in attribute `.#packages.x86_64-linux.qemu-example` where that option value is not defined.

This can be fixed in either/and two ways;

1. Add `default = {};` to your option definition (microvm/options.nix:606)
2. Add a fallback value to the end of the calculation statement ` ( credentialFiles or {} ); `

My personal sugestion is to apply 1 to preserve intent and prevent footguns around attribute usage. There is however a semantic difference between [unset] and [default value], to be complete, each requiring different approaches to test the value behind the option.

Ramblurr: Thanks @Bert-Proesmans for the comprehensive explanation.

SuperSandro2000:
> The deadnix workflow can't pass. It is configured to expect the source branch and the target branch in the same repo (this repo), but the branch is in my fork.

I hope I fixed that on main. If you rebase or in the next PR it should be fixed.

Also we cuaght a merge conflict that needs rebasing.

checks/default.nix
```diff
@@ -115,6 +115,37 @@ let
...
+      id = null;
+    } {
+      id = "credentials";
+      modules = [ ({ config, pkgs, ... }: {
+        # This is the guest vm config
+        microvm.credentialFiles.SECRET_BOOTSTRAP_KEY = "/etc/microvm-bootstrap.secret";
+        microvm.testing.enableTest = builtins.elem config.microvm.hypervisor [
+          # Hypervisors that support systemd credentials
+          "qemu"
+        ];
```
SuperSandro2000:
```suggestion
        microvm = {
          credentialFiles.SECRET_BOOTSTRAP_KEY = "/etc/microvm-bootstrap.secret";
          testing.enableTest = builtins.elem config.microvm.hypervisor [
            # Hypervisors that support systemd credentials
            "qemu"
          ];
        };
```

SuperSandro2000 (commented):
otherwise LGTM

Ramblurr:
force pushed to resolve merge conflicts, and also resolved the feedback above.

PR #339, opened 2025-02-25
fix: use target architecture binaries for disk volume creation

jpruiz84:
Use pkgs instead of pkgs.buildPackages to ensure e2fsprogs and other utilities are built for the target ARM64 architecture rather than the build x86 architecture. This resolves the "Exec format error" when running chattr and mkfs.ext4.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -55,7 +55,7 @@ rec {
               then lib.escapeShellArgs mkfsExtraArgs
               else " ";
       in (lib.optionalString autoCreate ''
-      PATH=$PATH:${with pkgs.buildPackages; lib.makeBinPath [ coreutils util-linux e2fsprogs xfsprogs dosfstools btrfs-progs ]}
+      PATH=$PATH:${with pkgs; lib.makeBinPath [ coreutils util-linux e2fsprogs xfsprogs dosfstools btrfs-progs ]}
 
       if [ ! -e '${image}' ]; then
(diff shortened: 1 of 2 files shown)


astro:
@mikatammi Please review

astro:
I fear that this is going to break the use-case where aarch64 is emulated by qemu on x86_64. Makes me wonder if we can reach a middleground that works for everyone by using `pkgs.hostPackages`?

avnik:
Probably no.  `runCommand` statements should use `pkgs.buildPackages` (on builder), runtime scripts should use `hostPackages` (ie for target platform). I intentionally keep emulation off on my main x86-64 machine, to catch cross-compilation bugs.

I would test, if it correctly cross-builds now

astro:
Well, in that case a runtime check for the host arch has to be done to select the right packages.

mikatammi:
image
Never forget!

mikatammi:
Yes I am bisexual and I have enjoyed sex with other men. If that is a bit too large to swallow for you, then maybe don't hire me into your projects!

mikatammi:
You cannot expect me to do code review for people, who kicked me out of their project because I put the PRIDE flag as my Slack profile picture. Yes I am bisexual, and I have enjoyed sex with other men 🖕

avnik:
@astro I checked it about week ago, and it at least builds for me!

PR #342, opened 2025-03-06
Allow to discard all toplevel references when using erofs

### LINKED ISSUE #228: Break storeDisk dependencies
A squashfs/erofs storeDisk is self-contained. It should not cause additional store paths to be required for a `nix copy`.

SuperSandro2000:
Closes #228

I kinda went and removed dead code on every file I opened while debugging.

### DIFF
--- a/lib/runner.nix
+++ b/lib/runner.nix
@@ -6,7 +6,7 @@
 let
   inherit (pkgs) lib;
 
-  inherit (microvmConfig) hostName virtiofsdScripts;
+  inherit (microvmConfig) hostName;
 
   inherit (import ./. { inherit lib; }) createVolumesScript makeMacvtap;
   inherit (makeMacvtap {
--- a/lib/runner.nix
+++ b/lib/runner.nix
@@ -72,7 +72,9 @@ pkgs.buildPackages.runCommand "microvm-${microvmConfig.hypervisor}-${hostName}"
   '') (builtins.attrNames binScriptPkgs)}
 
   mkdir -p $out/share/microvm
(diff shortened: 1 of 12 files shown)


lib/runner.nix
```diff
@@ -72,7 +72,9 @@ pkgs.buildPackages.runCommand "microvm-${microvmConfig.hypervisor}-${hostName}"
   '') (builtins.attrNames binScriptPkgs)}
 
   mkdir -p $out/share/microvm
+  ${lib.optionalString microvmConfig.systemSymlink ''
   ln -s ${toplevel} $out/share/microvm/system
+  ''}
```
SuperSandro2000, line 77: I first had this like that 

```suggestion
  '' + lib.optionalString microvmConfig.systemSymlink ''
  ln -s ${toplevel} $out/share/microvm/system
  '' + ''
```
which caused first an infinite recursion and after removing that code I got `9p/virtiofs shares not implemented for Firecracker` 😂  because probably the strings are tried to be added to the output of runCommand, not the multiline string itself.

nixos-modules/microvm/store-disk.nix
```diff
@@ -65,13 +65,15 @@ in
           pkgs.buildPackages.time
           pkgs.buildPackages.bubblewrap
           {
-            squashfs = [ pkgs.buildPackages.squashfs-tools-ng ];
```
SuperSandro2000, line 68: This only worked before because flattening nested lists like that produces an empty string. With __structuredAttrs it no longer does that.

astro:
Very awesome. Thank you!

PR #348, opened 2025-03-26
Change TimeoutStopSec= to TimeoutSec=

Dummyc0m:
A previous PR #231 ("Increased microvm default timeout stop to 2m30s") changed the default TimeoutStopSec= to 150 to avoid the VMs almost finish booting, but TimeoutStopSec=[1] is the amount of time systemd waits for the service to stop, not start. Switch to TimeoutSec= to set both TimeoutStartSec= and TimeoutStopSec=.

[1] https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#TimeoutStopSec=


astro:
Right, thanks!

Cc: @jpds

PR #349, opened 2025-03-29
feat: add evaluatedConfig field

FlafyDev:
This field allows using custom system configs, not necessarily nixos' eval-config

### DIFF
--- a/nixos-modules/host/options.nix
+++ b/nixos-modules/host/options.nix
@@ -25,6 +25,15 @@
     vms = mkOption {
       type = with types; attrsOf (submodule ({ config, name, ... }: {
         options = {
+          evaluatedConfig = mkOption {
+            description = ''
+              The evaluated configuration of this MicroVM, as a NixOS
+              module, for building **with** a flake.
+            '';
+            default = null;
+            type = nullOr types.unspecified;
+          };
+
           config = mkOption {
             description = ''
               A specification of the desired configuration of this MicroVM,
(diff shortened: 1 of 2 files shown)


nixos-modules/host/options.nix
```diff
@@ -25,6 +25,15 @@
     vms = mkOption {
       type = with types; attrsOf (submodule ({ config, name, ... }: {
         options = {
+          evaluatedConfig = mkOption {
+            description = ''
+              The evaluated configuration of this MicroVM, as a NixOS
+              module, for building **with** a flake.
```
astro: Is this for using `self.nixosConfigurations.*` that is usually already evaluated by `nixpkgs.lib.nixosSystem`? Could you perhaps clarify in the `description`?

FlafyDev: That is not the reason I need this option.

I'm using my own evaluation function for my config (https://github.com/flafydev/Combined-Manager). With this option I can use my function instead of having to use the default NixOS config eval function.

What you're saying can also be a good use case for this option.

What do you think of this description?
```
An already evaluated configuration of this MicroVM.
Allows supplying an already evaluated configuration or an alternative configuration evaluation function instead of NixOS's default eval-config.
```

FlafyDev: bump

astro: Yeah, that's more clear.

FlafyDev: Sorry I forgot about this. Done

SuperSandro2000 (approved):
LGTM

Issue #317, opened 2024-12-31
TimeoutStartUSec too low, or MicroVMs taking a little while be considered to start


pwaller:
I have a machine with (less than 10) microvms on it. When the machine is rebooted daily, the log looks like:

```
Dec 31 09:33:22 foo systemd[1]: Starting MicroVM 'machine-d'...
Dec 31 09:33:22 foo systemd[1]: Starting MicroVM 'machine-b'...
Dec 31 09:33:22 foo systemd[1]: Starting MicroVM 'machine-e'...
Dec 31 09:33:22 foo systemd[1]: Starting MicroVM 'machine-a'...
Dec 31 09:33:22 foo systemd[1]: Starting MicroVM 'machine-f'...
Dec 31 09:33:22 foo systemd[1]: Starting MicroVM machine-c'...
Dec 31 09:34:05 foo systemd[1]: Started MicroVM 'machine-a'.
Dec 31 09:34:52 foo systemd[1]: Started MicroVM machine-c'.
Dec 31 09:34:53 foo systemd[1]: microvm@machine-b.service: Failed with result 'timeout'.
Dec 31 09:34:53 foo systemd[1]: Failed to start MicroVM 'machine-b'.
Dec 31 09:34:53 foo systemd[1]: microvm@machine-d.service: Failed with result 'timeout'.
Dec 31 09:34:53 foo systemd[1]: Failed to start MicroVM 'machine-d'.
Dec 31 09:34:53 foo systemd[1]: microvm@machine-f.service: Failed with result 'timeout'.
Dec 31 09:34:53 foo systemd[1]: Failed to start MicroVM 'machine-f'.
Dec 31 09:34:53 foo systemd[1]: microvm@machine-e.service: Failed with result 'timeout'.
Dec 31 09:34:53 foo systemd[1]: Failed to start MicroVM 'machine-e'.
Dec 31 09:35:00 foo systemd[1]: Starting MicroVM 'machine-b'...
Dec 31 09:35:00 foo systemd[1]: Starting MicroVM 'machine-f'...
Dec 31 09:35:00 foo systemd[1]: Starting MicroVM 'machine-d'...
Dec 31 09:35:00 foo systemd[1]: Starting MicroVM 'machine-e'...
Dec 31 09:35:44 foo systemd[1]: Started MicroVM 'machine-e'.
Dec 31 09:35:44 foo systemd[1]: Started MicroVM 'machine-f'.
[...]

PR #353, opened 2025-04-24
Add option to configure TimeoutSec for vms

GlennWSo:
should resolve issue #317 

Make the default timeout for vm starup an option

### DIFF
--- a/nixos-modules/host/options.nix
+++ b/nixos-modules/host/options.nix
@@ -11,7 +11,7 @@
     };
     host.startupTimeout = mkOption {
       description = "Maximum allowed startup time (secs) for the vms";
-      type = int;
+      type = types.int;
       default = 150;
     };
 
--- a/nixos-modules/host/options.nix
+++ b/nixos-modules/host/options.nix
@@ -9,6 +9,11 @@
         Whether to enable the microvm.nix host module.
       '';
     };
+    host.startupTimeout = mkOption {
+      description = "Maximum allowed startup time (secs) for the vms";
+      type = int;
+      default = 150;
+    };
 
     host.useNotifySockets = mkOption {
       type = types.bool;
@@ -112,6 +117,7 @@
             type = bool;
             default = true;
           };
+
 
           restartIfChanged = mkOption {
             type = types.bool;
--- a/nixos-modules/host/options.nix
+++ b/nixos-modules/host/options.nix
@@ -10,8 +10,8 @@
       '';
     };
     host.startupTimeout = mkOption {
-      description = "Maximum allowed startup time (secs) for the vms";
-      type = types.int;
+      description = "Start up timeout for the VMs in seconds";
+      type = types.ints.positive;
       default = 150;
     };
 
--- a/nixos-modules/host/options.nix
+++ b/nixos-modules/host/options.nix
@@ -118,7 +118,6 @@
             default = true;
           };
 
-
           restartIfChanged = mkOption {
             type = types.bool;
             default = config.config != null;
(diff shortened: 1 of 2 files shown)


GlennWSo:
Changed to draft, because i have not yet confirmed if this will resolve the issue

astro:
While the option is sensible, the default of 90s is already quite high. VMs *should* boot within a minute. However, performance has degraded sometime in the last year and we're still figuring out the reason.

GlennWSo:
Yeah, this patchset does feel like treating the symptom not the cause.

In my case the VM has booted, but I can login with ssh. 

 But it restarts anyways because nextcloud-setup.service can take a long time to finnish when it needs to stuff like reindexing a database.  
 
 It seems to me that the deeper issue is that vmrunner does not differentiate between "booted" and "all oneshoot services on startup being completed"

GlennWSo:
I am using this as test systemds
to provoke the unwanted reboot.
```nix
# in the guest vm conf
  systemd.services."expensive-oneshot" = {
    enable = true;
    after = ["postgresql.service"];
    requires = ["postgresql.service"];
    before = ["long-derp.service"];
    wantedBy = ["multi-user.target"];
    script = ''
      #!/usr/bin/env bash
      echo running dummy script
      for i in {0..200}
      do
        sleep 1
        echo "Number: $i of 200"
      done

    '';
    serviceConfig = {
      Type = "oneshot";
    };
  };

  systemd.services."long-derp" = {
    enable = true;
    wantedBy = ["multi-user.target"];
    script = ''
      #!/bin/bash
      while true :
      do
      echo "Dont mind me"
      sleep 1
      done
    '';
    serviceConfig = {
      Type = "simple";
    };
  };
```

GlennWSo:
So i conculude if 

> I am using this as test systemds to provoke the unwanted reboot.
> 
> ```nix
> # in the guest vm conf
>   systemd.services."expensive-oneshot" = {
>     enable = true;
>     after = ["postgresql.service"];
>     requires = ["postgresql.service"];
>     before = ["long-derp.service"];
>     wantedBy = ["multi-user.target"];
>     script = ''
>       #!/usr/bin/env bash
>       echo running dummy script
>       for i in {0..200}
>       do
>         sleep 1
>         echo "Number: $i of 200"
>       done
> 
>     '';
>     serviceConfig = {
>       Type = "oneshot";
>     };
>   };
> 
>   systemd.services."long-derp" = {
>     enable = true;
>     wantedBy = ["multi-user.target"];
>     script = ''
>       #!/bin/bash
>       while true :
>       do
>       echo "Dont mind me"
>       sleep 1
>       done
>     '';
>     serviceConfig = {
>       Type = "simple";
>     };
>   };
> ```

With the default timeout setting this vm configuration gets stuck in endless rebooting. I assume this is because all services wanted multi-user.target is not reached in time.

But using the new proposed setting, I can give my vms enough time to complete my dummy setup service. 


```nix
 microvm = {

    host.startupTimeout = 500;

}
```

I have tested and confirmed that this solves #317 I will mark this pr as ready for merge.

Do you want me to write some test, or is "trust me bro" good enough?

nixos-modules/host/options.nix
```diff
@@ -113,6 +118,7 @@
             default = true;
           };
 
+
```
```suggestion
```

nixos-modules/host/options.nix
```diff
@@ -9,6 +9,11 @@
         Whether to enable the microvm.nix host module.
       '';
     };
+    host.startupTimeout = mkOption {
+      description = "Maximum allowed startup time (secs) for the vms";
```
```suggestion
      description = "Start up timeout for the VMs in seconds";
```

nixos-modules/host/options.nix
```diff
@@ -9,6 +9,11 @@
...
+      type = types.int;
```
```suggestion
      type = types.ints.positive;
```

SuperSandro2000:
To speed things up, I added some review comments :)

Issue #350, opened 2025-04-08
Support wildcards in PCI Device Identifiers for passthrough


avnik:
## Summary

It would be useful to support wildcards (such as `*`, `?`, or even regex-style patterns) in PCI device identifiers passed through to virtual machines using `microvm.nix`.

## Current Behavior

Currently, PCI devices must be specified with full, explicit addresses (e.g., `0000:00:1f.2`). This can be cumbersome in systems where PCI addresses are unstable across boots, or when dealing with multiple similar devices.

## Proposed Behavior

Allow wildcard patterns in the `path` field of PCI device definitions under `devices`.

### Example

```nix
{
  devices = [
    {
      bus = "pci";
      path = "0000:01:00.0";
    }
    {
      bus = "pci";
      path = "0000:01:*.*";
      deviceExtraArgs = "id=hostId,x-igd-opregion=on";
    }
  ];
}
```

In this example, the second device entry will match any PCI device under `0000:01:` and expand to all matching devices at **runtime**.

## Benefits

- Allows flexible, hardware-agnostic configurations.
- Avoids hardcoding device addresses that may change after hardware reordering or BIOS/firmware updates.
- Simplifies maintenance of reusable or dynamic VM configurations.
- Useful for systems with multiple similar devices, or unstable PCI enumeration.

## Implementation Considerations

- Wildcard resolution should happen at **runtime**, not during Nix evaluation.
- A helper script or systemd pre-start hook can be used to resolve patterns to actual PCI addresses before QEMU is launched.
- If multiple devices match, the system should either:
  - Apply the same options to all,
  - Or allow grouping/policy to control what gets passed.
- Validation/logging should be available to debug resolved devices.

## Related Work

- libvirt supports device selection by vendor/product IDs and other attributes.
- Similar device pattern matching is common in udev rules and systemd.

---

## Alternative / Experimental Workaround

A runtime workaround using inline shell substitution inside the Nix config has been explored, though it's not suitable for upstream inclusion due to its side effects and quoting issues. It may still be a useful reference or inspire a more integrated solution.

```nix
{
  path = "$(${pkgs.pciutils}/bin/lspci -Dnn | ${pkgs.gnugrep}/bin/grep ${d.vendorId}:${d.productId} | ${pkgs.coreutils}/bin/cut -c 1-12 | ${pkgs.coreutils}/bin/head -n 1)";
}
```

### Wrapper script for QEMU

[...]

PR #356, opened 2025-05-05
Moving pci/usb ids out of shell escaping

avnik:
This is an experimental attempt to implement #350 -- a dynamic PCI device id for passthrought.

Actual change move forming part of qemu command line outside of common shell escaping, replace it with simple concatenation. This allow use something like  `vfio-pci,host=$(/path/to/detect-id.sh)`.

Weak sides:
* Currently only qemu implemented
* Very barely tested
* possible security implications (why it _was_ under escaping)
* formatting probably wrong (`nix fmt` didn't work, what you use for format)

This is (not yet) a final solution, but PoC and point for a discussion

### DIFF
--- a/lib/runners/cloud-hypervisor.nix
+++ b/lib/runners/cloud-hypervisor.nix
@@ -213,14 +213,14 @@ in {
         })
         else throw "Unsupported interface type ${type} for Cloud-Hypervisor"
       ) interfaces)
-      ++
-      arg "--device" (map ({ bus, path, ... }: {
+    ) 
+    + " " + # Move vfio-pci outside of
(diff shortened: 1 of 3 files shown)


astro:
It's a pragmatic solution to this problem so I think I'll accept it. Could you please do the same for the other hypervisors?

mjm:
This is breaking for me, I believe because the function passed to `concatMapStringsSep` is returning a list instead of a string in the case of QEMU and crosvm. I think cloud-hypervisor probably works but those other two don't.

avnik:
It still be WIP, I planned to do more tests on the week

Issue #154, opened 2023-10-18
aarch64-darwin support?


LucaAttimis:
We would like to use `microvm` to provide staging and build environments for developers, but a few of them use apple hardware/software.

I can't find a way to use this project to run `aarch64-linux` vm (qemu) guest on `aarch64-darwin`, and I guess this is not supported.

...Is there any chance of supporting this? I am not *that* familiar with nix, but I might contribute a patch if you have a few pointer on where to start

astro:
I guess qemu should run but I don't run Darwin myself. Contributions welcome!

sgeop:
I also would love to have this capability on nix-darwin. I typically have used nixos or fedora in the past, and my work made me switch to macos on an m2 processor, which makes virtualization/containers really painful. Lima has been a decent solution, but a project this could be a game changer for mac / nix-darwin users.

Does anyone with knowledge of this project and of macos/darwin know how feasible this would be to implent?

ss-twieland:
Looking at the source code of `microvm`, I think it should be doable.

Not knowing anything about differences in CPU architectures and also not knowing anything about virtualisation packages (e.g. qemu) and their differences, I'm just pulling that statement out of thin air.

I started by adding "aarch64-darwin" to https://github.com/astro/microvm.nix/blob/main/flake.nix#L22.

This is a search result of `x86_64`:

```
pkgs/alioth.nix
29:    platforms = [ "aarch64-linux" "x86_64-linux" ];

flake-template/flake.nix
16:      system = "x86_64-linux";

lib/runners/cloud-hypervisor.nix
12:    x86_64-linux = "${kernel.dev}/vmlinux";
17:    if pkgs.stdenv.system == "x86_64-linux"

lib/runners/crosvm.nix
31:    x86_64-linux = "${kernel.dev}/vmlinux";

nixos-modules/microvm/options.nix
423:        Get a full list with `qemu-system-x86_64 -M help`
491:      lib.mkIf (pkgs.system == "x86_64-linux") (

checks/default.nix
15:        guestSystem = if "${system}" == "x86_64-linux" then "aarch64-unknown-linux-gnu"
16:          else "x86_64-linux";
24:            cpu = if "${system}" == "x86_64-linux" then "cortex-a53"

lib/runners/stratovirt.nix
24:    x86_64-linux =
32:    x86_64-linux = "ttyS0";

doc/src/host.md
26:      system = "x86_64-linux";

lib/runners/qemu.nix
[...]

PR #373, opened 2025-07-09
WIP: Try adding darwin support

podocarp:
The goal is to allow darwin hosts to build and run linux microvms #154 

The main problem is that with a "regular" cross compilation setup, you only have 2 entities in play: the host and target. So suppose you want to build a aarch64-linux image on a x86-linux machine.

But now there's no way for mac to build linux images, so we need to always run the builds in a VM usually, with either nix-darwin's linux-builder, or nix-rosetta-builder etc. This confuses the build though, since it thinks it is running in linux. In this case the host and target pkgs don't help (in fact they are probably both the same set, aarch64-linux for instance). This is what I was trying to achieve with the vmHostPackages option. 

The general strategy here is
- You need to run nix-darwin and configure linux-builder properly. This starts a linux vm locally that is used as a "remote" builder.
- Added an option `microvm.vmHostPackages` to override the `pkgs` used. Now, we set this to the aarch64-darwin pkgs, and this allows us to build aarch64-linux microVMs on aarch64-linux darwin builders but still be able to run it on the actual darwin host, since qemu and other utils are taken from vmHostPackages instead. 
- Some linux only flags in QEMU have been gated behind checks on `vmHostPackages` system. For example hvf vs kvm etc.

Some caveats
- I have only investigated QEMU. For sure the others won't work.
- xfs doesn't work on mac. Although I have fixed the create volume script dependencies to only pull in the filesystem utils that are needed, perhaps should mention in docs or add a check later on.
- virtiofs also doesn't seem to work on mac, so if you're doing the `ro-share` it has to be `9p`.
- not sure what the numa flags were for, but they certainly don't work on mac.
- the VM seems to be quite slow even with `accel=hvf` on M1, I will try later on another machine.


lib/runner.nix
```diff
@@ -27,15 +27,16 @@ let
     ''-a "microvm@${hostName}"'';
 
   vmHostPackages =
-    if microvmConfig.cpu == null
+    if microvmConfig.vmHostPackages != null then microvmConfig.vmHostPackages else
+    (if microvmConfig.cpu == null
```
astro: No need to parenthesize here

lib/default.nix
```diff
@@ -32,7 +44,7 @@ rec {
   createVolumesScript = pkgs: volumes:
     lib.optionalString (volumes != []) (
       lib.optionalString (lib.any (v: v.autoCreate) volumes) ''
-        PATH=$PATH:${with pkgs.buildPackages; lib.makeBinPath [ coreutils util-linux e2fsprogs xfsprogs dosfstools btrfs-progs ]}
+        PATH=$PATH:${with pkgs.buildPackages; lib.makeBinPath ([ coreutils ] ++ (collectFsUtils volumes pkgs))}
```
astro: I think it's no longer clear where `pkgs` comes from. We should get rid of the `with` clause.

lib/default.nix
```diff
@@ -1,4 +1,16 @@
 { lib }:
+let
+  fsTypeToUtil = pkgs: fs: with pkgs;{
```
astro: The `pkgs` here is `buildPackages`? Otherwise, we could use the global instance.

lib/default.nix
```diff
@@ -1,4 +1,16 @@
...
+    "ext2" = e2fsprogs;
+    "ext3" = e2fsprogs;
+    "ext4" = e2fsprogs;
+    "xfs"= xfsprogs;
+    "btrfs" = btrfs-progs;
+    "vfat" = dosfstools;
+  }.${fs};
+  collectFsTypes = volumes: map (v: v.fsType) volumes;
+  collectFsUtils = volumes: pkgs: map (fsType: fsTypeToUtil pkgs fsType) (collectFsTypes volumes);
```
astro: `pkgs:` comes first in the other functions.

nixos-modules/microvm/options.nix
```diff
@@ -483,6 +483,13 @@ in
       '';
     };
 
+    vmHostPackages = mkOption {
+      description = "If set, overrides the default host package.";
```
astro, line 488: Excuse me for not knowing how a Mac works, but how exactly is this going to be used? Is that really enough documentation to be obvious all nix-darwin users? :-)

astro (commented):
Could somebody with Darwin please review?

lib/default.nix
podocarp: No, and the naming is a bit unfortunate... createVolumeScript takes in a `pkgs` parameter and then passes it around, so this `pkgs` is actually whatever createVolumeScript received.

podocarp: There's a few ways around this
1. We keep this
2. We can make vmHostPackages an extra parameter to lib/default.nix and pass it in. But this gets messy (I tried) because there are quite a few call sites, and not all of them require vmHostPackages
3. We could just set vmHostPackages with a default value of the `host.cpu == null ...` check, and access it from `config` whenever we need it. I think this is the cleanest solution.

nixos-modules/microvm/options.nix
podocarp: Yes the docs are quite lacking... I'm going to need to do a more detailed writeup

lib/runner.nix
```diff
@@ -6,15 +6,16 @@
...
-  inherit (import ./. { inherit lib; }) createVolumesScript makeMacvtap withDriveLetters;
+  inherit (import ./. { inherit lib; }) makeMacvtap makeMacvtap withDriveLetters;
+  inherit (import ./volumes.nix { inherit lib microvmConfig; }) createVolumesScript;
   inherit (makeMacvtap {
     inherit microvmConfig hypervisorConfig;
   }) openMacvtapFds macvtapFds;
 
   hypervisorConfig = import (./runners + "/${microvmConfig.hypervisor}.nix") {
-    inherit pkgs microvmConfig macvtapFds withDriveLetters;
+    inherit microvmConfig macvtapFds pkgs;
```
SuperSandro2000, line 18:
```suggestion
    inherit pkgs microvmConfig macvtapFds withDriveLetters;
```
I think you did a bad merge here

lib/volumes.nix
```diff
@@ -0,0 +1,79 @@
...
+  }) defaultFsType;
+
+  fsTypeToUtil = fs: with vmHostPackages; {
+      "ext2" = e2fsprogs;
+      "ext3" = e2fsprogs;
+      "ext4" = e2fsprogs;
+      "xfs" = xfsprogs;
+      "btrfs" = btrfs-progs;
+      "vfat" = dosfstools;
+    }.${fs};
```
SuperSandro2000, line 15:
```suggestion
    }.${fs} or (throw "Do not know how to handle ${fs}");
```

lib/volumes.nix
```diff
@@ -0,0 +1,79 @@
...
+                      "-n"
+                    else
+                      (lib.warnIf (label != null)
+                        "Will not label volume ${label} with filesystem type ${fsType}. Open an issue on the microvm.nix project to request a fix."
+                        null
+                      )
+                  )
+                else
+                  null;
+              labelArgument = if (labelOption != null && label != null) then "${labelOption} '${label}'" else "";
```
SuperSandro2000, line 63:
```suggestion
              labelArgument = lib.optionalString (labelOption != null && label != null) "${labelOption} '${label}'";
```

SuperSandro2000 (commented):
This PR is rather big and the commit history not super clean.

Could we extract some changes from this PR to make it smaller? Maybe we can extract some bits or abstraction or general cross fix to make this easier to review?

podocarp:
Yea sorry I guess I was mostly just playing around with it. I'll redo it -- first, I think we can introduce an override for host packages for users for special conditions such as this. Then, we add some checks for darwin. I'm closing this and redoing, which will take some time because I'm on vacation

PR #375, opened 2025-07-09
Make crosvm use virtiofsd

nesteroff:
Crosvm requires the virtiofs backend to support VHOST_USER_PROTOCOL_F_CONFIG which is enabled by passing --tag to virtiofsd.

Using virtiofsd aligns crosvm with other VMMs and resolves issues with the writable /nix/store overlay.

### DIFF
--- a/lib/runners/crosvm.nix
+++ b/lib/runners/crosvm.nix
@@ -97,16 +97,14 @@ in {
         ]
       ) volumes
       ++
-      builtins.concatMap ({ proto, tag, source, ... }:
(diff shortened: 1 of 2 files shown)


astro:
Built-in virtiofs in crosvm is convenient, making me wonder if we should make it configurable? Are you going to try to get more vhost-* devices usable?

I don't use crosvm myself a lot so I would also merge as-is.

astro:
The permission problem is reason enough. Thank you!

PR #377, opened 2025-07-16
cloud-hypervisor: allow overriding `--serial` and `--console`

### LINKED ISSUE #376: cloud-hypervisor: cannot override the `--serial` and `--console` args.
### Issue

The cloud-hypervisor runner unconditionally sets `--serial` and `--console`.

https://github.com/astro/microvm.nix/blob/f30ae9b9e740116d044935c9c0c68bd1c2ce8432/lib/runners/cloud-hypervisor.nix#L142-L143

### Motivation

...but there are other values that could be useful (at least for myself)

```
      --console <console>
          Control (virtio) console: "off|null|pty|tty|file=</path/to/a/file>,iommu=on|off" [default: tty]
      --serial <serial>
          Control serial port: off|null|pty|tty|file=</path/to/a/file>|socket=</path/to/a/file> [default: null]
```

Also see the slow boot problem caused by the default `--serial tty` in https://github.com/astro/microvm.nix/issues/366

### Current behavior

If you try to set `--serial` for example, you get the error:

```
$ nix run .#packages.x86_64-linux.cloud-hypervisor-example --show-trace
warning: Git tree '/home/ramblurr/src/microvm.nix' is dirty
error: the argument '--serial <serial>' cannot be used multiple times

Usage: microvm@cloud-hypervisor-microvm [OPTIONS]

For more information, try '--help'.
```

### Possible solutions

I understand the goal not to have an additional nixos module option for every single CLI flag for the VMMs, that would just get out of hand.

I think a good solution is to detect if the user passed a flag in the `extraArgs` list, and if so, do not add the default that microvm otherwise would.

This is easy to implement now using the [`extractOptValues`](https://github.com/astro/microvm.nix/pull/336/files#diff-84f980eccc1d3f99c10f8b0e6c5c5fc2af2069517533c4c4d73264a190e85188R81-R135) function I implemented over in #336 

~~If you approve this, then I can prepare a PR that does exactly that.~~

A PR for this is over in #377 so you can see exactly what this looks like.

Ramblurr:
cloud-hypervisor: allow overriding --serial and --console arguments

Check if --serial or --console are provided in extraArgs before adding
defaults. This allows users to specify custom values like file output
or socket connections instead of being forced to use the hardcoded
defaults. We skip the default kernel serial parameter when serial is not
set to tty.

Fixes #376

Builds on and requires ~PR #336~

### DIFF
--- a/lib/runners/cloud-hypervisor.nix
+++ b/lib/runners/cloud-hypervisor.nix
@@ -12,7 +12,7 @@ let
 
   hasUserConsole = (extractOptValues "--console" extraArgs).values != [];
   hasUserSerial = (extractOptValues "--serial" extraArgs).values != [];
-  userSerial = if hasUserSerial then (extractOptValues "--serial" extraArgs).values else "";
+  userSerial = lib.optionalString hasUserSerial (extractOptValues "--serial" extraArgs).values;
 
   kernelPath = {
     x86_64-linux = "${kernel.dev}/vmlinux";
@@ -26,7 +26,7 @@ let
     then "console=ttyAMA0"
     else "";
 
-  kernelConsole = if (!hasUserSerial || userSerial == "tty") then kernelConsoleDefault else "";
+  kernelConsole = lib.optionalString (!hasUserSerial || userSerial == "tty") kernelConsoleDefault;
 
   kernelCmdLine = "${kernelConsole} reboot=t panic=-1 ${builtins.unsafeDiscardStringContext (toString microvmConfig.kernelParams)}";
 


lib/runners/cloud-hypervisor.nix
```diff
@@ -10,18 +10,26 @@ let
...
 
-  kernelConsole =
+  kernelConsoleDefault =
     if pkgs.stdenv.system == "x86_64-linux"
     then "earlyprintk=ttyS0 console=ttyS0"
     else if pkgs.stdenv.system == "aarch64-linux"
     then "console=ttyAMA0"
     else "";
 
+  kernelConsole = if (!hasUserSerial || userSerial == "tty") then kernelConsoleDefault else "";
```
```suggestion
  kernelConsole = lib.optionalString (!hasUserSerial || userSerial == "tty") kernelConsoleDefault;
```

lib/runners/cloud-hypervisor.nix
```diff
@@ -10,18 +10,26 @@ let
   inherit (microvmConfig) vcpu mem balloon initialBalloonMem deflateOnOOM hotplugMem hotpluggedMem user interfaces volumes shares socket devices hugepageMem graphics storeDisk storeOnDisk kernel initrdPath;
   inherit (microvmConfig.cloud-hypervisor) platformOEMStrings extraArgs;
 
+  hasUserConsole = (extractOptValues "--console" extraArgs).values != [];
+  hasUserSerial = (extractOptValues "--serial" extraArgs).values != [];
+  userSerial = if hasUserSerial then (extractOptValues "--serial" extraArgs).values else "";
```
```suggestion
  userSerial = lib.optionalString hasUserSerial (extractOptValues "--serial" extraArgs).values;
```

Issue #378, opened 2025-07-16
cloud-hypervisor: Problems running multiple microvms because cid is hardcoded


Ramblurr:
# Problem 

The microvm cloud-hypervisor runner is hardcoding the cid for vsock at `3`.

That isn't good because:
- cids should be unique per vm
- it makes the `microvm.vsock.cid` useless for chv

https://github.com/astro/microvm.nix/blob/f30ae9b9e740116d044935c9c0c68bd1c2ce8432/lib/runners/cloud-hypervisor.nix#L153

I suppose hardcoded vsock was added to support systemd-notify, but it ignores the user's configuration and causes problems when running multiple VMs from running on the same host.

## Solution

The fix is to:
1. Respect `microvm.vsock.cid` when set
2. Extract and merge any user-provided `--vsock` options from `extraArgs`
3. Only use the hardcoded default when neither is specified
4. Throw a clear error if both `microvm.vsock.cid` and `--vsock cid=...` are provided

This allows users to either:
- Set `microvm.vsock.cid = 5;` in their config
- Pass `--vsock "socket=/custom/path.sock"` via extraArgs
- Get the default behavior if they don't configure vsock

The implementation properly handles partial user options (e.g., just cid or just socket) and merges them with sensible defaults, while maintaining backward compatibility for users who don't configure vsock at all.

I guess this is a breaking change for users who were relying on the automatic systemd-notify support, but I'm not sure that will really affect runtime?

We could add a text in the release like: 

> Previously, cloud-hypervisor would always configure vsock with `cid=3` and enable systemd-notify notifications automatically. With this change, users must explicitly set `microvm.vsock.cid` to enable the notify feature.
> 
> To maintain the previous behavior, users should add to their configuration:
> 
>     microvm.vsock.cid = 3;  # or any unique CID for your VM

PR #379, opened 2025-07-16
cloud-hypervisor: respect `microvm.vsock.cid` instead of hardcoding

Ramblurr:
cloud-hypervisor: respect `microvm.vsock.cid` instead of hardcoding

Previously, cloud-hypervisor always configured vsock with cid=3, causing
problems when runnning multiple VMs from running on the same host due to CID
conflicts. This change respects `microvm.vsock.cid` when set and allows users to
configure vsock via `microvm.cloud-hypervisor.extraArgs`.

The implementation:
- Uses `microvm.vsock.cid` when specified
- Extracts and merges `--vsock` options from extraArgs
- Throws an error if both vsock.cid and --vsock cid=... are provided
- Warns users when vsock is not configured (disabling systemd-notify)

Fixes: #378

### DIFF
--- a/lib/runners/cloud-hypervisor.nix
+++ b/lib/runners/cloud-hypervisor.nix
@@ -43,7 +43,7 @@ let
   userVSockPath = extractParamValue "socket" userVSockStr;
   userVSockCID = extractParamValue "cid" userVSockStr;
   vsockCID = if vsock.cid != null && userVSockCID != null
-             then throw "Cannot set `microvm.vsock.cid` and --vsock 'cid=${userVSockCID}...' from `microvm.cloud-hypervisor.extraArgs` at the same time"
+             then throw "Cannot set `microvm.vsock.cid` and --vsock 'cid=${userVSockCID}...' via `microvm.cloud-hypervisor.extraArgs` at the same time"
              else if vsock.cid != null
                   then vsock.cid
                   else userVSockCID;
@@ -131,7 +131,7 @@ let
   platformExtracted = extractOptValues "--platform" extraArgs;
   extraArgsWithoutPlatform = platformExtracted.args;
   userPlatformOpts = platformExtracted.values;
-  userPlatformStr = if userPlatformOpts == [] then "" else builtins.head userPlatformOpts;
+  userPlatformStr = lib.optionalString (userPlatformOpts != []) (builtins.head userPlatformOpts);
   userHasOemStrings = (extractParamValue "oem_strings" userPlatformStr) != null;
   platformOps =
     if userHasOemStrings then


Ramblurr:
As I said in matrix:  [Astro (@astro:envs.net)](https://matrix.to/#/@astro:envs.net): I have a series of patches for the cloud-hypervisor runner stacking up, most of them depend on [PR #336](https://github.com/astro/microvm.nix/pull/336), because of a handy function I added to help parse out args from extraArgs so we can merge with them. Github's ui is problematic when it comes to dependent PRs like this.. just let me know if you want me to put them all in a single PR or want me to rebase an individual one... they each have self contained commits for a clean history

Ramblurr:
@SuperSandro2000 This one is rebased and ready for review :)

lib/runners/cloud-hypervisor.nix
```diff
@@ -101,14 +126,18 @@ let
...
-  supportsNotifySocket = true;
-
   oemStringValues = platformOEMStrings ++ lib.optional supportsNotifySocket "io.systemd.credential:vmm.notify_socket=vsock-stream:2:8888";
   oemStringOptions = lib.optional (oemStringValues != []) "oem_strings=[${lib.concatStringsSep "," oemStringValues}]";
   platformExtracted = extractOptValues "--platform" extraArgs;
   extraArgsWithoutPlatform = platformExtracted.args;
   userPlatformOpts = platformExtracted.values;
-  platformOps = lib.concatStringsSep "," (oemStringOptions ++ userPlatformOpts);
+  userPlatformStr = if userPlatformOpts == [] then "" else builtins.head userPlatformOpts;
+  userHasOemStrings = (extractParamValue "oem_strings" userPlatformStr) != null;
```
```suggestion
  userHasOemStrings = (extractParamValue "oem_strings" (toString userPlatformOpts)) != null;
```
Does that work?

lib/runners/cloud-hypervisor.nix
```diff
@@ -30,6 +37,24 @@ let
 
   kernelCmdLine = "${kernelConsole} reboot=t panic=-1 ${builtins.unsafeDiscardStringContext (toString microvmConfig.kernelParams)}";
 
+
+  userVSockOpts = (extractOptValues "--vsock" extraArgs).values;
+  userVSockStr = if userVSockOpts == [] then null else builtins.head userVSockOpts;
+  userVSockPath = extractParamValue "socket" userVSockStr;
+  userVSockCID = extractParamValue "cid" userVSockStr;
```
SuperSandro2000, line 44:
```suggestion
  userVSockPath = extractParamValue "socket" (toString userVSockOpts);
  userVSockCID = extractParamValue "cid" (toString userVSockOpts);
```
same as below

lib/runners/cloud-hypervisor.nix
```diff
@@ -30,6 +37,24 @@ let
...
+  vsockCID = if vsock.cid != null && userVSockCID != null
+             then throw "Cannot set vsock.cid and --vsock 'cid=${userVSockCID}...' at the same time"
```
SuperSandro2000: Can we mention the options from which those came from?

lib/runners/cloud-hypervisor.nix
```diff
@@ -30,6 +37,24 @@ let
...
+             else if vsock.cid != null
+                  then vsock.cid
+                  else userVSockCID;
+  supportsNotifySocket = vsockCID != null;
+  vsockPath = if userVSockPath != null then userVSockPath else "notify.vsock";
```
SuperSandro2000, line 51: Maybe we should move that in the option default to have that better documented? Or would that be overwritten by existing set options?

lib/runners/cloud-hypervisor.nix
Ramblurr: Hm, this relies on the fact that `toString [] == ""` and `toString ["some-val"] == "some-val"`, my `nix repl` says this works but I think the original is more clear. Up to you.

lib/runners/cloud-hypervisor.nix
Ramblurr: There is currently no option where one can set the vsock path. I would leave it as is for now, because I have an upcoming PR that better implements the notify socket support feature (also for qemu) and can address this with a `microvm.vsock.path` option.

Ramblurr:
@SuperSandro2000 I pushed a commit fixing some feedback, for the other feedback I left comments. Let me know your decision and I will happily change.

lib/runners/cloud-hypervisor.nix
SuperSandro2000: Also true

lib/runners/cloud-hypervisor.nix
```diff
@@ -101,14 +126,18 @@ let
...
 
-  supportsNotifySocket = true;
-
   oemStringValues = platformOEMStrings ++ lib.optional supportsNotifySocket "io.systemd.credential:vmm.notify_socket=vsock-stream:2:8888";
   oemStringOptions = lib.optional (oemStringValues != []) "oem_strings=[${lib.concatStringsSep "," oemStringValues}]";
   platformExtracted = extractOptValues "--platform" extraArgs;
   extraArgsWithoutPlatform = platformExtracted.args;
   userPlatformOpts = platformExtracted.values;
-  platformOps = lib.concatStringsSep "," (oemStringOptions ++ userPlatformOpts);
+  userPlatformStr = if userPlatformOpts == [] then "" else builtins.head userPlatformOpts;
```
```suggestion
  userPlatformStr = lib.optionalString (userPlatformOpts != []) builtins.head userPlatformOpts;
```

lib/runners/cloud-hypervisor.nix
```diff
@@ -30,6 +37,24 @@ let
 
   kernelCmdLine = "${kernelConsole} reboot=t panic=-1 ${builtins.unsafeDiscardStringContext (toString microvmConfig.kernelParams)}";
 
+
+  userVSockOpts = (extractOptValues "--vsock" extraArgs).values;
+  userVSockStr = if userVSockOpts == [] then null else builtins.head userVSockOpts;
+  userVSockPath = extractParamValue "socket" userVSockStr;
+  userVSockCID = extractParamValue "cid" userVSockStr;
+  vsockCID = if vsock.cid != null && userVSockCID != null
+             then throw "Cannot set `microvm.vsock.cid` and --vsock 'cid=${userVSockCID}...' from `microvm.cloud-hypervisor.extraArgs` at the same time"
```
```suggestion
             then throw "Cannot set `microvm.vsock.cid` and --vsock 'cid=${userVSockCID}...' via `microvm.cloud-hypervisor.extraArgs` at the same time"
```

lib/runners/cloud-hypervisor.nix
SuperSandro2000: I'll leave this open so that we can find it again easily but fine for me.

Ramblurr:
merge conflict fixed and changes incorporated :)

lib/runners/cloud-hypervisor.nix
```diff
@@ -30,6 +37,24 @@ let
 
   kernelCmdLine = "${kernelConsole} reboot=t panic=-1 ${builtins.unsafeDiscardStringContext (toString microvmConfig.kernelParams)}";
 
+
+  userVSockOpts = (extractOptValues "--vsock" extraArgs).values;
```
astro, line 41: Why do it like this instead of getting these values from `microvmConfig`?

Ramblurr: `microvm.vsock` does not expose all possible vsock option, nor should it I think.


```
      --vsock <vsock>
          Virtio VSOCK parameters "cid=<context_id>,socket=<socket_path>,iommu=on|off,id=<device_id>,pci_segment=<segment_id>"
```

but the user might pass CHV specific `--vsock` options in `microvm.cloud-hypervisor.extraArgs`, so this code ensures that their options are taken into account.

astro: I see.

Still, I prefer structural options over inviting the user to use stringly configuration that we then parse.

Ramblurr: I'm happy to implement all the chv block options under microvm.cloud-hypervisor.vsock of you want. I don't think it's worth it though. I only added the --vsock parsing here because I think escape hatches are important. `extraArgs` is fundamentally an escape hatch. How do you want to proceed?

SuperSandro2000: If we implement all possible arguments as options, we might easily run into problems when trying to support NixOS stable and unstable.

We should probably improve the structured config in the next PRs to make them better and reduce the need for extraArgs but right now I think this is the best thing we can do without giant effort.

SuperSandro2000: I talked astro soft: for my sake

So until we have better options this is what we have and if you have some time to improve the typed options that would be really appreciated!

PR #381, opened 2025-07-17
Simplify qemu shutdown script?

podocarp:
https://qemu-project.gitlab.io/qemu/interop/qemu-qmp-ref.html#command-QMP-machine.system_powerdown

I'm assuming this is doing an ACPI shutdown request based on how they describe it. But that should be good enough right? For me the current keystroke sequence doesn't even work.

Also I can't help but to modify the close by balloon script as well
- Don't do anything if balloon is off
- No need to query capabilities if tail is just gonna strip it away... user can't see it anyways
- Throw if socket isn't defined as well

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -331,7 +331,7 @@ lib.warnIf (mem == 2048) ''
     then
       ''
         (
-          ${writeQmp { execute = "qmp_capabilities"; }}
+          ${writeQmp { execute = "query-commands"; }}
           ${writeQmp { execute = "system_powerdown"; }}
           # wait for exit
           cat
@@ -341,19 +341,20 @@ lib.warnIf (mem == 2048) ''
     else throw "Cannot shutdown without socket";
 
   setBalloonScript =
-    if socket != null && balloon
-    then ''
+    if balloon
+    then
+      if socket != null then
+      ''
       VALUE=$(( $SIZE * 1024 * 1024 ))
       SIZE=$( (
-        ${writeQmp { execute = "qmp_capabilities"; }}
         ${writeQmp { execute = "balloon"; arguments.value = 987; }}
       ) | sed -e s/987/$VALUE/ | \
         ${pkgs.socat}/bin/socat STDIO UNIX:${socket},shut-none | \
-        tail -n 1 | \
         ${pkgs.jq}/bin/jq -r .data.actual \
       )
       echo $(( $SIZE / 1024 / 1024 ))
-    ''
+      ''
+      else throw "Cannot balloon without socket"
     else null;
 
   requiresMacvtapAsFds = true;


podocarp:
Possible extension: we could do a `sleep x` in the script and send another kill request to force shutdown? This seems to just be an acpi shutdown which doesn't work if the guest is stuck or not booted up yet.

astro:
> Possible extension: we could do a sleep x in the script and send another kill request to force shutdown? This seems to just be an acpi shutdown which doesn't work if the guest is stuck or not booted up yet.

When run in a systemd service, systemd will kill after timeout already.

lib/runners/qemu.nix
```diff
@@ -332,46 +332,16 @@ lib.warnIf (mem == 2048) ''
...
+          # wait for exit
           cat
         ) | \
         ${pkgs.socat}/bin/socat STDIO UNIX:${socket},shut-none
     ''
     else throw "Cannot shutdown without socket";
 
   setBalloonScript =
-    if socket != null
+    if socket != null && balloon
```
SuperSandro2000: This line shoud be changed in the other commit which probably should also be an extra commit.

lib/runners/qemu.nix
```diff
@@ -331,7 +331,7 @@ lib.warnIf (mem == 2048) ''
     then
       ''
         (
-          ${writeQmp { execute = "qmp_capabilities"; }}
+          ${writeQmp { execute = "query-commands"; }}
```
SuperSandro2000, line 334: What is your motivation to change this?

lib/runners/qemu.nix
```diff
@@ -341,19 +341,20 @@ lib.warnIf (mem == 2048) ''
     else throw "Cannot shutdown without socket";
 
   setBalloonScript =
-    if socket != null && balloon
-    then ''
+    if balloon
+    then
+      if socket != null then
+      ''
```
SuperSandro2000, line 347: If we change this again, we can drop the change in the last commit, to make the history cleaner.

lib/runners/qemu.nix
```diff
@@ -341,19 +341,20 @@ lib.warnIf (mem == 2048) ''
...
       VALUE=$(( $SIZE * 1024 * 1024 ))
       SIZE=$( (
-        ${writeQmp { execute = "qmp_capabilities"; }}
         ${writeQmp { execute = "balloon"; arguments.value = 987; }}
       ) | sed -e s/987/$VALUE/ | \
         ${pkgs.socat}/bin/socat STDIO UNIX:${socket},shut-none | \
-        tail -n 1 | \
```
SuperSandro2000, line 382: How did you test that this did not break anything?

podocarp: That's because I removed the echo capabilities line, which doesn't do anything because it gets tailed away. That line doesn't seem to do anything at all because the user won't be able to see the output.

podocarp: I am sorry I just read the docs; I did not realise you had to execute capabilities to establish a connection, what a strange requirement

lib/runners/qemu.nix
podocarp: Same as below, I did not know this was required.

PR #387, opened 2025-07-19
Add option to pass cpu-config to firecracker

wtfaremyinitials:
Add new option to pass custom [cpu-config](https://github.com/firecracker-microvm/firecracker/blob/7dfe2765849c2444a22defa11be8641508c5ce5c/docs/cpu_templates/cpu-templates.md#custom-cpu-templates) JSON to firecracker.

### DIFF
--- a/lib/runners/firecracker.nix
+++ b/lib/runners/firecracker.nix
@@ -11,11 +11,14 @@ let
     interfaces volumes shares devices
     kernel initrdPath
     storeDisk;
+  inherit (microvmConfig.firecracker) cpuConfig;
 
   kernelPath = {
     x86_64-linux = "${kernel.dev}/vmlinux";
     aarch64-linux = "${kernel.out}/${pkgs.stdenv.hostPlatform.linux-kernel.target}";
   }.${system};
+
+  cpu-config = lib.optionalAttrs (cpuConfig != null) pkgs.writeText "cpu-config.json" (builtins.toJSON cpuConfig);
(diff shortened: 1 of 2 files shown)


lib/runners/firecracker.nix
```diff
@@ -11,12 +11,15 @@ let
     interfaces volumes shares devices
     kernel initrdPath
     storeDisk;
+  inherit (microvmConfig.firecracker) cpuConfig;
```
astro: I think `config` is rather redundant in an option name. It's called `cpu` for qemu, too (albeit it has a different type).

astro: Also, the variable is only used in one line. Move the `let ... in` more locally.

lib/runners/firecracker.nix
```diff
@@ -11,12 +11,15 @@ let
...
 
   kernelPath = {
     x86_64-linux = "${kernel.dev}/vmlinux";
     aarch64-linux = "${kernel.out}/${pkgs.stdenv.hostPlatform.linux-kernel.target}";
   }.${system};
 
+  cpu-config = lib.optionalAttrs (cpuConfig != null) pkgs.writeText "cpu-config.json" (builtins.toJSON cpuConfig);
```
astro: Convention would be to use `cpuConfig` case.

astro (changes requested):
Thank you!

PR #393, opened 2025-07-25
Allow passing extra arguments to the VMM at runtime

nesteroff:
This adds an option to pass extra arguments to the VMM at runtime via a script. It can be used to dynamically gather system information or detect devices that need to be passed to the VM before launching.

### DIFF
--- a/lib/runner.nix
+++ b/lib/runner.nix
@@ -43,8 +43,11 @@ let
       ${preStart}
       ${createVolumesScript vmHostPackages microvmConfig.volumes}
       ${lib.optionalString (hypervisorConfig.requiresMacvtapAsFds or false) openMacvtapFds}
+      runtime_args=${lib.optionalString (microvmConfig.extraArgsScript != null) ''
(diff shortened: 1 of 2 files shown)


lib/runner.nix
```diff
@@ -40,11 +40,12 @@ let
   binScripts = microvmConfig.binScripts // {
     microvm-run = ''
       set -eou pipefail
+      unset EXTRA_ARGS
```
SuperSandro2000: Why do we need to unset EXTRA_ARGS here?

Also we should probably document this somewhere.

nesteroff: > Why do we need to unset EXTRA_ARGS here?

We unset it just in case a variable with the same name was set earlier somewhere else so it doesn’t mess up the runner command.

> Also we should probably document this somewhere.

Sure, I updated the option description.

nesteroff:
Added a separate option for this to make the implementation a bit cleaner.

PR #400, opened 2025-08-20
tap-up: execute after network.target

MarcelCoding:
required bridge devices may not exists until `network.target` is reached


MarcelCoding:
I thought I found a bug, but I could not test this yet, because our hypervisor was not rebooted yet.

PR #421, opened 2025-10-19
chore: replace hostName with fqdnOrHostName

felbinger:
Using fully qualified domain names as unique identifier for containers improves identification of containers on large (potentially multi tenant) hypervisors.

In the current version using a container name containing a dot doesn't work. I forked microvm.nix and adjusted it so that fqdn's are used by default.

Should we try to upstream this? If yes how?

I drafted this change because merging this would result in breaking changes if `networking.domain` is being set for a machine.

### DIFF
--- a/lib/runner.nix
+++ b/lib/runner.nix
@@ -6,7 +6,7 @@
 let
   inherit (pkgs) lib;
 
-  inherit (microvmConfig) hostName vmHostPackages;
+  inherit (microvmConfig) fqdnOrHostName vmHostPackages;
 
   inherit (import ./. { inherit lib; }) makeMacvtap withDriveLetters extractOptValues extractParamValue;
   inherit (import ./volumes.nix { pkgs = microvmConfig.vmHostPackages; }) createVolumesScript;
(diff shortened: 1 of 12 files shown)


SuperSandro2000 (commented):
> Modules that accept a mere networking.hostName but prefer a fully qualified domain name may use networking.fqdnOrHostName instead.

https://search.nixos.org/options?channel=unstable&show=networking.fqdn&query=networking.fqdn

felbinger:
> I haven't had the time to test yet, will do in the next days...

Ok works. Now this change works without breaking changes if `networking.domain` is NOT set.

If I am correct, systems that have a domain set will still encounter breaking changes, right?

How should we proceed?

SuperSandro2000:
@astro how do we could handle this?

SuperSandro2000:
Also the CI tests are failing for some reason I did not grasp yet.

nixos-modules/host/default.nix
```diff
@@ -66,7 +66,7 @@ in
         description = "Install MicroVM '${name}'";
         before = [
           "microvm@${name}.service"
-          "microvm-tap-interfaces@${name}.service"
+          "microvm-tap-interfaces@${name}.service" # TODO split after first .
```
astro: Splitting will lead to conflicts between `vm.example.com` and `vm.example.org` which I assumed this PR was going to solve. I guess we should rather [escape](https://www.man7.org/linux/man-pages/man1/systemd-escape.1.html) here.

astro (changes requested):
Is this a breaking change only if `networking.domain` is defined?

nixos-modules/host/default.nix
felbinger: Can you elaborate on this conflict? In my tests I couldn't find any issues with it

PR #422, opened 2025-10-22
Allow using microvm.interfaces.*.type = "vmnet-shared" for addressable darwin networking

esselius:
As `tap`/`macvtap` is a bit tricky on darwin and `user` networking doesn't create a routable IP + port forwarding doesn't seem to work with `hvf` virt accel.


astro:
This is just for qemu?

Would you mind adding a description in `doc/src/interfaces.md`?

PR #424, opened 2025-10-29
Directory permissions for ${stateDir} are set wrongly on the initial creation of it

0xCCF4:
Problem: On the first usage of microvm in host mode, on my machine the state directory is created with `rw-r rwx r-x` permissions. This results in the systemd units failing with `Changing to the requested working directory failed: Permission denied` since the user `microvm` is not allowed to chdir to the `${stateDir}`

One possible solution: Change the permission of the `${stateDir}` to `775` during activation.

### DIFF
--- a/nixos-modules/host/default.nix
+++ b/nixos-modules/host/default.nix
@@ -27,7 +27,7 @@ in
     system.activationScripts.microvm-host = ''
       mkdir -p ${stateDir}
       chown ${user}:${group} ${stateDir}
-      chmod 775 ${stateDir}
+      chmod u+rwx,g+w ${stateDir}
     '';
 
     environment.systemPackages = [


nixos-modules/host/default.nix
```diff
@@ -27,7 +27,7 @@ in
     system.activationScripts.microvm-host = ''
       mkdir -p ${stateDir}
       chown ${user}:${group} ${stateDir}
-      chmod g+w ${stateDir}
+      chmod 775 ${stateDir}
```
SuperSandro2000: Not sure if we would want 770 here 🤔

0xCCF4: As far as I understand, 7xx is required to run the systemd services, which run under the user owning the folder.

Since I am not familiar with the project's source. I assumed that x7x is the desired configuration for the ${stateDir} - inferred from the chmod g+w.

Setting xx0 would be properly ok, otherwise, do you think that `chmod u+rwx,g+w` would do the trick? Letting the end-user decide on the others' permission?

astro: If that's arbitrary, how about making it configurable with a NixOS option along with a sensible default?

SuperSandro2000: > Setting xx0 would be properly ok, otherwise, do you think that `chmod u+rwx,g+w` would do the trick? Letting the end-user decide on the others' permission?

Sounds good to me and making it configurable may also be a good idea but could also be to detailed.

0xCCF4: > If that's arbitrary, how about making it configurable with a NixOS option along with a sensible default?

The others' permission is arbitrary, yes. Creating a new option for that could be an idea, but I believe that no new option would be required to control that if we went with  `chmod u+rwx,g+w`, since setting the permission that way would only enforce that the microvm services are working. A user who wants to control the other permissions on the file can simply imperatively change it or write their own script that modifies the file permissions.

PR #427, opened 2025-11-06
firecracker: Set --enable-pci option by default when available

bjackman:
This enables the PCI virtio transport which is dramatically faster than the MMIO one.

Support for this flag is currently only availble in Nixpkgs unstable.

However I'm not aware of any reason for _not_ setting this when it's available, so when it is, set it by default.

### DIFF
--- a/nixos-modules/microvm/options.nix
+++ b/nixos-modules/microvm/options.nix
@@ -590,6 +590,12 @@ in
       description = "Custom CPU template passed to firecracker.";
     };
 
+    firecracker.enablePci = mkOption {
+      type = types.bool;
+      default = lib.versionOlder "1.13.0" pkgs.firecracker.version;
+      description = "Use PCI Virtio transport, improves performance.";
+    };
+
     prettyProcnames = mkOption {
       type = types.bool;
       default = true;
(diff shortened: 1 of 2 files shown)


nixos-modules/microvm/options.nix
```diff
@@ -590,6 +590,12 @@ in
       description = "Custom CPU template passed to firecracker.";
     };
 
+    firecracker.enablePci = mkOption {
+      type = types.bool;
+      default = lib.versionOlder "1.13.0" pkgs.firecracker.version;
```
bjackman, line 595: Wait what, this is backwards. Why did this work for me?

bjackman: Lol, I used the wrong function , but I also got the args the wrong way around so the two bugs cancelled each other out.

bjackman: Wait, but also, this whole thing is garbage. `enablePci` doesn't exist, I have no idea why this seemed to be working for me, I must have got my `--override-input` args wrong or something.

bjackman:
Closing this until I can clean up my garbage, I am no longer seeing any perf improvements after I fixed it, I think I was testing some random noise

PR #428, opened 2025-11-06
firecracker: Set --enable-pci option by default when available

bjackman:
This enables the PCI virtio transport which is supposed to be faster than the MMIO one.

Support for this flag is currently only availble in Nixpkgs unstable.

I haven't been able to measure these perf improvements.

However this is still what the Firecracker docs say to do, so I guess let's make it the default?

### DIFF
--- a/lib/runners/firecracker.nix
+++ b/lib/runners/firecracker.nix
@@ -86,7 +86,7 @@ in {
     then throw "hotpluggedMem not implemented for Firecracker"
     else if credentialFiles != {}
     then throw "credentialFiles are not implemented for Firecracker"
-    else lib.escapeShellArgs [
+    else lib.escapeShellArgs ([
       "${pkgs.firecracker}/bin/firecracker"
       "--config-file" configFile
       "--api-sock" (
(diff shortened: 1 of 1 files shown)


lib/runners/firecracker.nix
```diff
@@ -86,15 +86,15 @@ in {
...
+    else lib.escapeShellArgs ([
       "${pkgs.firecracker}/bin/firecracker"
       "--config-file" configFile
       "--api-sock" (
         if socket != null
         then socket
         else throw "Firecracker must be configured with an API socket (option microvm.socket)!"
       )
-    ];
+    ] ++ lib.optional microvmConfig.firecracker.enablePci "--enable-pci");
```
astro: Does it make a lot of sense to have an option to enable PCI? How about doing just:

```suggestion
    ] ++ lib.optional (lib.versionAtLeast pkgs.firecracker.version "1.13.0") "--enable-pci");
```

bjackman: Yeah I dunno - I certainly don't need the option myself. Let's drop it, it could always be added back if there's a need later.

PR #429, opened 2025-11-06
firecracker: Increase serial baudrate

bjackman:
By default the kernel uses a slow baudrate which, at least on my host, slows down boot by about 10 seconds.

### DIFF
--- a/lib/runners/firecracker.nix
+++ b/lib/runners/firecracker.nix
@@ -24,7 +24,7 @@ let
     boot-source = {
       kernel_image_path = kernelPath;
       initrd_path = initrdPath;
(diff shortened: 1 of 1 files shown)


lib/runners/firecracker.nix
```diff
@@ -24,7 +24,7 @@ let
     boot-source = {
       kernel_image_path = kernelPath;
       initrd_path = initrdPath;
-      boot_args = "console=ttyS0 noapic acpi=off reboot=k panic=1 i8042.noaux i8042.nomux i8042.nopnp i8042.dumbkbd ${toString microvmConfig.kernelParams}";
```
bjackman, line 27: ugh

PR #432, opened 2025-11-16
feat: implement `microvm.vms.*.extraModules`

### LINKED ISSUE #197: Pass additional NixOS modules to VM
I use custom NixOS modules to make additional NixOS options available to my servers, and I would like to do the same for VMs. However, I could not find a way to do this. I am using the fully declarative MicroVM setup.

An option like the following would be awesome:
```nix
{ microvm, ... }: {
  imports = [ microvm.host ];
  microvm.vms = {
    my-microvm = {

      extraModules = [ ];

      config = {
        # [...]
      };
    };
  };
}
```

ajaxbits:
Adds an option at `microvm.vms.*.extraModules` which allows users to pass in a list of NixOS Modules that will be merged into `microvm.vms.*.config`, enabling fully-declarative MicroVM setups to use existing NixOS Modules from other projects.

This is achieved by passing the `extraModules` option through as [an argument to `lib/eval-config.nix`](https://github.com/NixOS/nixpkgs/blob/256e2076ea1aac921f970927c58e1737072cf463/nixos/lib/eval-config.nix#L33C1-L33C22).

Fixes #197.

### DIFF
--- a/nixos-modules/host/options.nix
+++ b/nixos-modules/host/options.nix
@@ -52,7 +52,7 @@
                     ../microvm
                   ] ++ (map (x: x.value) defs);
                 prefix = [ "microvm" "vms" name "config" ];
-                inherit (config) specialArgs pkgs;
+                inherit (config) extraModules specialArgs pkgs;
                 system =
                   if config.pkgs != null then
                     config.pkgs.stdenv.hostPlatform.system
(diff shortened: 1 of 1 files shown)


nixos-modules/host/options.nix
```diff
@@ -102,6 +102,23 @@
             '';
           };
 
+          extraModules = mkOption {
+            type = types.listOf types.deferredModule;
```
ajaxbits, line 106: Seeking feedback on the typing of this option. `deferredModule` seemed correct, but I am not a SME on module definitions like these.

SuperSandro2000: Not sure if https://github.com/NixOS/nixpkgs/blob/a421ac6595024edcfbb1ef950a3712b89161c359/nixos/modules/misc/documentation.nix#L307 is more correct

ajaxbits: Nice find -- I could totally see that being the way to go. I'll let astro weigh in if they have an opinion before I change anything

SuperSandro2000:
I think astro refers to something like:

```
{ microvm, ... }: {
  imports = [ microvm.host ];
  microvm.vms = {
    my-microvm = {
      config = {
        imports = [ HERE ];
      };
    };
  };
}
```

I also think that would work but we could be missing something here.

astro:
@ajaxbits Thank you for persisting. I acknowledge that `nixos.extraModules` seems to be not available outside documentation, and `imports` doesn't work if you ever access `config`. Therefore this makes sense.

ajaxbits:
Thank you for all your hard work on this project @astro!

PR #434, opened 2025-11-18
extraConfig and extraArgs for firecracker

bjackman:
```
❯❯  git log -n 2
commit 2c0e70ee0c7d36902e6f2fd6d523dedfb5e51846 (HEAD -> firecracker-extraconfig, fork/firecracker-extraconfig)
Author: Brendan Jackman <jackmanb@google.com>
Date:   Mon Nov 10 15:20:49 2025 +0000

    fireracker: Add extraArgs option
    
    For experimentation, support adding arbitrary extra args.

commit b6c5f9c5008463f0769e4c6b70aa0a26c98c9fbb
Author: Brendan Jackman <jackmanb@google.com>
Date:   Fri Nov 7 13:42:11 2025 +0000

    firecracker: Add extraConfig option
    
    For experimentation with low-level settings, add support for merging in
    arbitrary options to the Firecracker config file, following the standard pattern
    in nixpkgs for this kind of thing.
    
    My immediate usecase for this is for experimenting with firecracker features
    that are not yet released, so it doesn't make sense to add more specific Nix
    options for those features.
```

### DIFF
--- a/nixos-modules/microvm/options.nix
+++ b/nixos-modules/microvm/options.nix
@@ -603,7 +603,9 @@ in
     };
 
     firecracker.extraConfig = mkOption {
-      type = types.attrs;
+      type = types.submodule {
+        freeformType = (pkgs.formats.json {}).type;
+      };
       default = {};
       description = "Extra config to merge into Firecracker JSON configuration";
     };


nixos-modules/microvm/options.nix
```diff
@@ -590,6 +590,18 @@ in
...
     };
 
+    firecracker.extraArgs = mkOption {
+      type = with types; listOf str;
+      default = [];
+      description = "Extra arguments to pass to firecracker.";
+    };
+
+    firecracker.extraConfig = mkOption {
+      type = types.attrs;
```
SuperSandro2000: That type does not merge. We need to use a submodule of attrs.

PR #436, opened 2025-11-21
fix: don't copy nixpkgs to the store needlessly

### LINKED ISSUE #435: bug: needless copy of nixpkgs
This line here:
https://github.com/microvm-nix/microvm.nix/blob/c4e4a264da114c618251b17eb4c959f86376e530/nixos-modules/host/options.nix#L41

Causes a needless recopying of nixpkgs to the store, even if it is already in the store. This is due to nix's behavior when using a path in interpolation. It automatically imports any interpolated paths to the /nix/store (strange I know, but that's what it does).

it's possible simply changing to:
`import (config.nixpkgs + "/nixos/lib/eval-config.nix")` would work, but I'll test a patch to make sure before submitting.

nrdxp:
if the value of nixpkgs is a path, it will be copied to the store by nix's implicit interpolation. Coercing it to a string beforehand should resolve this.

resolves #435

### DIFF
--- a/nixos-modules/host/options.nix
+++ b/nixos-modules/host/options.nix
@@ -38,7 +38,7 @@
             default = null;
             type = nullOr (lib.mkOptionType {
               name = "Toplevel NixOS config";
-              merge = loc: defs: (import "${config.nixpkgs}/nixos/lib/eval-config.nix" {
+              merge = loc: defs: (import "${toString config.nixpkgs}/nixos/lib/eval-config.nix" {
                 modules =
                   let
                     extraConfig = ({ lib, ... }: {


nrdxp:
testing evaluation, and nixpkgs is no longer being copied with this patch

nixos-modules/host/options.nix
```diff
@@ -38,7 +38,7 @@
             default = null;
             type = nullOr (lib.mkOptionType {
               name = "Toplevel NixOS config";
-              merge = loc: defs: (import "${config.nixpkgs}/nixos/lib/eval-config.nix" {
+              merge = loc: defs: (import "${toString config.nixpkgs}/nixos/lib/eval-config.nix" {
```
SuperSandro2000, line 41: Don't we just throw away the string context here?

nrdxp: path literals don't have string context, and no, converting them to a string does not "throw it away"

SuperSandro2000: Yeah, you're probably right

SuperSandro2000 (approved):

PR #440, opened 2025-12-07
fix: append dont clear

brianmcgillion:
lib.optional returns the empty set [] when a check is false. In this case it causes the options to be set to an empty list which the module does not allow. So append the new options if true.

issue introduced from (https://github.com/microvm-nix/microvm.nix/pull/419)

### DIFF
--- a/nixos-modules/microvm/mounts.nix
+++ b/nixos-modules/microvm/mounts.nix
@@ -87,7 +87,7 @@ lib.mkIf config.microvm.guest.enable {
           upperdir = "${writableStoreOverlay}/store";
           workdir = "${writableStoreOverlay}/work";
         };
-        options = lib.optional isRwStoreVirtiofsShare "userxattr";
+        options = lib.mkIf isRwStoreVirtiofsShare [ "userxattr" ];
       };
     }
   ) {


nixos-modules/microvm/mounts.nix
```diff
@@ -87,7 +87,8 @@ lib.mkIf config.microvm.guest.enable {
           upperdir = "${writableStoreOverlay}/store";
           workdir = "${writableStoreOverlay}/work";
         };
-        options = lib.optional isRwStoreVirtiofsShare "userxattr";
+      } // lib.optionalAttrs isRwStoreVirtiofsShare {
+        options = lib.mkAfter [ "userxattr" ];
```
SuperSandro2000: And why do we need the mkAfter?

SuperSandro2000 (commented):
Also I just wanted to let you know that we don't use semantic commits.

nixos-modules/microvm/mounts.nix
brianmcgillion: I was thinking the ordering could matter. Fixed

nixos-modules/microvm/mounts.nix
```diff
@@ -87,7 +87,8 @@ lib.mkIf config.microvm.guest.enable {
           upperdir = "${writableStoreOverlay}/store";
           workdir = "${writableStoreOverlay}/work";
         };
-        options = lib.optional isRwStoreVirtiofsShare "userxattr";
+      } // lib.optionalAttrs isRwStoreVirtiofsShare {
+        options = [ "userxattr" ];
```
SuperSandro2000:
```suggestion
        options = lib.mkIf isRwStoreVirtiofsShare [ "userxattr" ];
```
We should really use this 😅

brianmcgillion: fixed. sorry about this. It is always the one-liners that take the time ...

PR #442, opened 2025-12-08
Add support for creating PCIe root ports in QEMU

nesteroff:
This change adds the ability to define PCIe root ports for QEMU virtual machines. These ports can be used for device hot-plugging, allowing vfio-pci devices to be attached dynamically.

In the current version, it's possible to create ports via extraArgs and set them in deviceExtraArgs but with recent changes (https://github.com/microvm-nix/microvm.nix/pull/437), this approach produces "bus not found" errors. There is likely a race condition between creating vfio-pci devices which are now placed at the beginning of the QEMU command and the PCIe ports that were appended at the end.

With this change, PCIe root ports are explicitly created before the vfio-pci devices that may use them.

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -54,7 +54,7 @@ let
     else "threads";
 
   inherit (microvmConfig) hostName vcpu mem balloon initialBalloonMem deflateOnOOM hotplugMem hotpluggedMem user interfaces shares socket forwardPorts devices vsock graphics storeOnDisk kernel initrdPath storeDisk credentialFiles;
-  inherit (microvmConfig.qemu) machine extraArgs serialConsole;
+  inherit (microvmConfig.qemu) machine extraArgs serialConsole pcieRootPorts;
 
 
   volumes = withDriveLetters microvmConfig;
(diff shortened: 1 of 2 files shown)


lib/runners/qemu.nix
```diff
@@ -197,11 +197,20 @@ lib.warnIf (mem == 2048) ''
...
       "-device" "virtio-rng-${devType}"
-    ])
+    ] ++
+    # Create PCIe root ports before vfio-pci devices that might require them
+    builtins.concatMap ({ bus, id, chassis, ... }:
+      [ "-device"
+        "pcie-root-port,bus=${bus},id=${id},chassis=${toString chassis}"
+      ]
+    ) pcieRootPorts
+    )
```
astro, line 211: The lines above now require two spaces of indentation.

nixos-modules/microvm/options.nix
```diff
@@ -559,6 +575,44 @@ in
       '';
     };
 
+    qemu.pcieRootPorts = mkOption {
+      description = ''
+        A list of PCIe root ports that can be used for device hot-plugging or
```
astro: Interesting feature. Can we add a link that explains how hot-plugging works?

Could we even generate some helper scripts for the user based on this configuration?

brianmcgillion: hotplug main feature
https://github.com/tiiuae/vhotplug

and the downstream PR, that uses this patch
https://github.com/tiiuae/ghaf/pull/1624

overall we are quite heavy users of passthrough with microvm.
https://github.com/tiiuae/ghaf/blob/main/modules/hardware/common/devices.nix
https://github.com/tiiuae/ghaf/blob/main/modules/hardware/common/qemu.nix
https://github.com/tiiuae/ghaf/tree/main/modules/microvm
https://github.com/tiiuae/ghaf/tree/main/modules/hardware/passthrough

if there is appetite. I think that it could be great to move some of the functionality from Ghaf to microvm, so that it would be readily available to more folks.

nesteroff: > Interesting feature. Can we add a link that explains how hot-plugging works?

It works the same way as for other QEMU devices, simply by using the device_add and device_del commands. The problem is that on the Q35 machine type, the PCIe root bus does not support hot-plugging. To enable it, PCIe root ports must be created first. I updated the option description and included a link to the QEMU PCIe Guidelines for further reference.

nixos-modules/microvm/options.nix
```diff
@@ -559,6 +575,65 @@ in
...
+        For additional details see the QEMU PCI Express Guidelines:
+        <https://gitlab.com/qemu-project/qemu/-/blob/master/docs/pcie.txt>
+      '';
+      default = [];
+      example = literalExpression /* nix */ ''
+        [ {
+          bus = "pcie.0";
+          id = "pci_port_0";
+          chassis = 0;
+        }]
```
```suggestion
        } ]
```

nixos-modules/microvm/options.nix
```diff
@@ -559,6 +575,65 @@ in
...
+      '';
+      type = with types; listOf (submodule {
```
SuperSandro2000, line 596: If the entire type can be a list of null, then each individual option does not need to be nullable, unless each option can be set individually.

nesteroff: Thanks, good to know. I think it doesn’t really make sense to allow all parameters to be null because the ports wouldn’t be usable without id so I made it non-null.

nesteroff:
> Can you rebase on master to fix the eval/CI?

Sure, all green now.

SuperSandro2000 (approved):

PR #447, opened 2025-12-17
Remove external dependency on `flake-utils`

irgolic:
Hey, thanks for creating such a useful abstraction over VMs!

Would you be open to reducing the size of your dependency tree? `flake-utils` is only used for the `eachSystem` function. Inlining or replacing it cuts the project's lockfile in half (as `flake-utils` itself depends on another flake). 

All of these commits remove the external dependency in different ways; let me know if you would like me to amend or remove any of them:
- The first commit (57e124415ad00d4a229279da530c4de9bd815644) inline-vendors `forAllSystems`, including an MIT header. 
- The second commit (d5f82d04f34b0d4a568da6e3fe1eecd91f60250b) replaces it with a simpler function that retains parity. 
- The third commit (37af54cabdede37018ce26ddb34e63d0f4595084) removes the behavior of automatically including `currentSystem` if the flake is run with `--impure`. I'm guessing that most developers depending on `flake-utils` are unaware of this behavior – in my opinion explicitly relying on [`nix-systems`](https://github.com/nix-systems/nix-systems) for this behavior is a better way to achieve this goal.

While this PR aims to remove the external dependency on `flake-utils` in the simplest way possible, in the long term, I suggest [rewriting the flake with a `forAllSystems` function](https://ayats.org/blog/no-flake-utils#do-we-really-need-flake-utils) for better readability (you won't need to [merge two big attrsets](https://github.com/microvm-nix/microvm.nix/blob/5a18de0d3ff46971d2074c2326e45768b1a9de5f/flake.nix#L138), instead keeping it all top-level).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,23 +1,5 @@
 {
   "nodes": {
-    "flake-utils": {
-      "inputs": {
-        "systems": "systems"
-      },
-      "locked": {
-        "lastModified": 1731533236,
-        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
-        "type": "github"
-      },
-      "original": {
-        "owner": "numtide",
-        "repo": "flake-utils",
-        "type": "github"
-      }
-    },
     "nixpkgs": {
       "locked": {
         "lastModified": 1759381078,
@@ -36,7 +18,6 @@
     },
     "root": {
       "inputs": {
-        "flake-utils": "flake-utils",
         "nixpkgs": "nixpkgs",
         "spectrum": "spectrum"
       }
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
First of all the git history of this is not very at all nice. This should be one commit, as otherwise using git blame becomes needlessly way harder to use and we have lots of back and forth.

Then why not use lib.foldAttrs? That would make the code probably way shorter and easier to read.

You linked https://ayats.org/blog/no-flake-utils#do-we-really-need-flake-utils that is pretty short and if we touch that part of the code, we should go with that and do it completely.

> Inlining or replacing it cuts the project's lockfile in half (as `flake-utils` itself depends on another flake).

True, but that wouldn't matter to much in any flakes I know where microvm is used, as there are usually at least a few other flake-utils. But for people not using 20 other flakes, there is a small benefit.

SuperSandro2000 (changes requested):
see above comment

irgolic:
Thanks! Indeed, microvm-nix is currently the only flake I currently depend on that uses `flake-utils`.

I've changed this PR to consist of only one commit, implementing the pattern at https://ayats.org/blog/no-flake-utils#do-we-really-need-flake-utils. It turned out pretty clean, and removed a level of indentation from most of the file.

Let me know if there's anything else you'd like me to amend.

PR #449, opened 2025-12-23
Add options for providing hypervisor and virtiofsd packages

### LINKED ISSUE #270: package options for runners
Would it be ideal to expose a `package` option for runners to allow easier overriding of the package used?
For example I wish to use https://github.com/64kramsystem/qemu-pinning however this requires a full overlay right now to set this for microvms.

https://github.com/astro/microvm.nix/blob/69e8ac63a7a4d40bb65e73af41ba60df2eba0419/lib/runners/qemu.nix#L38C10-L38C22

elmankku:
This adds options for providing packages for selected hypervisors (QEMU, crosvm, firecracker and cloud-hypervisor) and virtiofsd. This makes it easier e.g. for having different versions for different VMs/builds, which would be a great help for maintaining more complex systems.

Resolves #270.

### DIFF
--- a/lib/runners/stratovirt.nix
+++ b/lib/runners/stratovirt.nix
@@ -41,8 +41,8 @@ let
     if requirePci
     then
       if addr < 32
-      then "pci,bus=pcie.0,addr=0x${pkgs.lib.toHexString addr}"
-      else throw "Too big PCI addr: ${pkgs.lib.toHexString addr}"
+      then "pci,bus=pcie.0,addr=0x${lib.toHexString addr}"
+      else throw "Too big PCI addr: ${lib.toHexString addr}"
     else "device";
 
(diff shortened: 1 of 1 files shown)


nixos-modules/microvm/options.nix
```diff
@@ -804,6 +836,13 @@ in
       '';
     };
 
+    virtiofsd.package = mkOption {
+      description = "The virtiofsd package to use.";
+      type = types.package;
+      default = cfg.vmHostPackages.virtiofsd;
+      defaultText = literalExpression ''cfg.vmHostPackages.virtiofsd'';
```
```suggestion
      defaultText = literalExpression ''config.microvm.vmHostPackages.virtiofsd'';
```

astro (approved):

PR #453, opened 2025-12-29
feat: add systemd-machined integration and SSH over VSOCK

### LINKED ISSUE #123: Enter running machine as systemd service
Is it possible to connect terminal stdin/stdout to deployed machine, to inspect what's going on there?

elsbrock:
Adds systemd-machined integration and SSH over VSOCK support for MicroVMs.

## machined integration

`microvm.registerWithMachined = true` registers VMs with machined:

```
$ machinectl list
MACHINE    CLASS SERVICE     OS VERSION ADDRESSES
authelia   vm    microvm.nix -  -       -
blackpearl vm    microvm.nix -  -       -
```

- `machinectl list/status/show` for visibility
- `machinectl terminate/kill` to stop VMs

This PR prefers `RegisterMachineEx` on systemd 259+ and falls back to legacy machined calls on older hosts.  
When available, it registers `VSockCID` / `SSHAddress` metadata (plus network interfaces).

`machinectl reboot/poweroff/terminate/kill` all stop the VM without auto-restart (systemd treats signals from machined as intentional stops). Use `systemctl restart microvm@<name>` for restarts.

## SSH over VSOCK

`microvm.vsock.ssh.enable = true` configures guest sshd to listen on VSOCK (requires `microvm.vsock.cid`):

```
$ microvm -s myvm
Connecting to myvm via VSOCK...
[myvm]$
```

Works with qemu/crosvm/kvmtool (native AF_VSOCK) and cloud-hypervisor (socket mux).

This PR also updates `microvm -s` argument handling so SSH options can be forwarded via `--`, e.g.:

- `microvm -s myvm -- -l root -i /path/to/key`

## Other changes

- `microvm -l` respects configured `microvm.stateDir`
- `microvm.machineId` option for consistent UUID handling across machined/SMBIOS/guest machine-id use
- QEMU shutdown script handles missing control socket (fixes restart after external kill)
- imperative `microvm@%i` template tolerates missing `microvm-register` / `microvm-unregister` scripts

## Tests

- expands machined test coverage for VM metadata (`VSockCID`, `SSHAddress`) where applicable
- adds a focused `microvm -s` argument-handling test
- adds imperative template test for start/stop without register scripts

Closes #123

### DIFF
--- a/checks/machined.nix
+++ b/checks/machined.nix
@@ -1,0 +1,58 @@
+{ self, nixpkgs, system, hypervisor }:
+
+let
+  pkgs = nixpkgs.legacyPackages.${system};
+  vmName = "machined-test";
+in
+{
+  # Test that MicroVMs can be registered with systemd-machined
+  "machined-${hypervisor}" = pkgs.nixosTest {
+    name = "machined-${hypervisor}";
+    nodes.host = { lib, ... }: {
+      imports = [ self.nixosModules.host ];
+
+      virtualisation.qemu.options = [
+        "-cpu"
+        {
+          "aarch64-linux" = "cortex-a72";
+          "x86_64-linux" = "kvm64,+svm,+vmx";
+        }.${system}
+      ];
+      virtualisation.diskSize = 4096;
+
+      # Define a VM with machined registration enabled
+      microvm.vms.${vmName}.config = {
+        microvm = {
+          hypervisor = hypervisor;
+          # Enable machined registration on the VM
+          registerWithMachined = true;
+        };
+        networking.hostName = vmName;
+        system.stateVersion = lib.trivial.release;
+      };
+    };
+    testScript = ''
+      # Wait for the MicroVM service to start
+      host.wait_for_unit("microvm@${vmName}.service", timeout = 1200)
+
+      # Verify the VM is registered with machined
+      host.succeed("machinectl list | grep -q '${vmName}'")
+
+      # Verify machine status works
+      host.succeed("machinectl status '${vmName}'")
+
+      # Verify the machine class is 'vm'
+      host.succeed("machinectl show '${vmName}' --property=Class | grep -q 'vm'")
+
+      # Verify leader PID exists
+      host.succeed("machinectl show '${vmName}' --property=Leader | grep -q '^Leader=[0-9]'")
+
+      # Terminate the VM via machinectl (sends SIGTERM to hypervisor)
+      host.succeed("machinectl terminate '${vmName}'")
+
+      # Wait for the service to stop
+      host.wait_until_fails("machinectl status '${vmName}'", timeout = 30)
+    '';
+    meta.timeout = 1800;
+  };
+}
--- a/checks/machined.nix
+++ b/checks/machined.nix
@@ -1,58 +1,80 @@
 { self, nixpkgs, system, hypervisor }:
 
 let
-  pkgs = nixpkgs.legacyPackages.${system};
   vmName = "machined-test";
+  vsockCid = 4242;
 in
 {
   # Test that MicroVMs can be registered with systemd-machined
-  "machined-${hypervisor}" = pkgs.nixosTest {
-    name = "machined-${hypervisor}";
-    nodes.host = { lib, ... }: {
-      imports = [ self.nixosModules.host ];
-
-      virtualisation.qemu.options = [
-        "-cpu"
+  "machined-${hypervisor}" =
+    import (nixpkgs + "/nixos/tests/make-test-python.nix")
+      (
+        { pkgs, lib, ... }:
         {
-          "aarch64-linux" = "cortex-a72";
-          "x86_64-linux" = "kvm64,+svm,+vmx";
-        }.${system}
-      ];
-      virtualisation.diskSize = 4096;
+          name = "machined-${hypervisor}";
+          nodes.host = {
+            imports = [ self.nixosModules.host ];
 
-      # Define a VM with machined registration enabled
-      microvm.vms.${vmName}.config = {
-        microvm = {
-          hypervisor = hypervisor;
-          # Enable machined registration on the VM
-          registerWithMachined = true;
-        };
-        networking.hostName = vmName;
-        system.stateVersion = lib.trivial.release;
+            boot.kernelModules = [ "kvm" ];
+
+            virtualisation.qemu.options = [
+              "-cpu"
+              {
+                "aarch64-linux" = "cortex-a72";
+                "x86_64-linux" = "kvm64,+svm,+vmx";
+              }
+              .${system}
+            ];
+            virtualisation.diskSize = 4096;
+
+            # Define a VM with machined registration enabled
+            microvm.vms.${vmName}.config = {
+              microvm = {
+                inherit hypervisor;
+                # Enable machined registration on the VM
+                registerWithMachined = true;
+              }
+              // lib.optionalAttrs (hypervisor != "cloud-hypervisor") {
+                vsock.cid = vsockCid;
+              };
+              networking.hostName = vmName;
+              system.stateVersion = lib.trivial.release;
+            };
+          };
+          testScript = ''
+            # Wait for the MicroVM service to start
+            host.wait_for_unit("microvm@${vmName}.service", timeout = 60)
+            # ^ this is actually none blocking
+
+            # Verify the VM is registered with machined
+            host.wait_until_succeeds("machinectl list | grep -q '${vmName}'", timeout=240)
+
+            # Verify machine status works
+            host.succeed("machinectl status '${vmName}'")
+
+            # Verify the machine class is 'vm'
+            host.succeed("machinectl show '${vmName}' --property=Class | grep -q 'vm'")
+
+            ${lib.optionalString ((lib.versionAtLeast pkgs.systemd.version "259") && hypervisor != "cloud-hypervisor") ''
+              # On systemd >=259 RegisterMachineEx path should expose VSOCK/SSH metadata
+              host.succeed("machinectl show '${vmName}' --property=VSockCID | grep -q 'VSockCID=${toString vsockCid}'")
+              host.succeed("machinectl show '${vmName}' --property=SSHAddress | grep -q 'SSHAddress=vsock/${toString vsockCid}'")
+            ''}
+
+            # Test kill command (send signal 0 to check process exists)
+            host.succeed("machinectl kill '${vmName}' --signal=0")
+
+            # Terminate the VM via machinectl (sends SIGTERM to hypervisor)
+            host.succeed("machinectl terminate '${vmName}'")
+
+            # Wait for the service to stop
+            host.wait_until_fails("machinectl status '${vmName}'", timeout = 30)
+          '';
+          meta.timeout = 600;
+        }
+      )
+      {
+        inherit system;
+        pkgs = nixpkgs.legacyPackages.${system};
       };
-    };
-    testScript = ''
-      # Wait for the MicroVM service to start
-      host.wait_for_unit("microvm@${vmName}.service", timeout = 1200)
-
-      # Verify the VM is registered with machined
-      host.succeed("machinectl list | grep -q '${vmName}'")
-
-      # Verify machine status works
-      host.succeed("machinectl status '${vmName}'")
-
-      # Verify the machine class is 'vm'
-      host.succeed("machinectl show '${vmName}' --property=Class | grep -q 'vm'")
-
-      # Verify leader PID exists
-      host.succeed("machinectl show '${vmName}' --property=Leader | grep -q '^Leader=[0-9]'")
-
-      # Terminate the VM via machinectl (sends SIGTERM to hypervisor)
-      host.succeed("machinectl terminate '${vmName}'")
-
-      # Wait for the service to stop
-      host.wait_until_fails("machinectl status '${vmName}'", timeout = 30)
-    '';
-    meta.timeout = 1800;
-  };
 }
--- a/checks/machined.nix
+++ b/checks/machined.nix
@@ -61,9 +61,6 @@ in
               host.succeed("machinectl show '${vmName}' --property=SSHAddress | grep -q 'SSHAddress=vsock/${toString vsockCid}'")
             ''}
 
-            # Test kill command (send signal 0 to check process exists)
-            host.succeed("machinectl kill '${vmName}' --signal=0")
-
             # Terminate the VM via machinectl (sends SIGTERM to hypervisor)
             host.succeed("machinectl terminate '${vmName}'")
 
--- a/lib/runner.nix
+++ b/lib/runner.nix
@@ -6,7 +6,7 @@
 let
   inherit (pkgs) lib;
 
-  inherit (microvmConfig) hostName machineId vmHostPackages;
+  inherit (microvmConfig) hostName vmHostPackages;
 
   inherit (import ./. { inherit lib; }) makeMacvtap withDriveLetters extractOptValues extractParamValue;
   inherit (import ./volumes.nix { pkgs = microvmConfig.vmHostPackages; }) createVolumesScript;
@@ -24,13 +24,23 @@ let
   tapMultiQueue = hypervisorConfig.tapMultiQueue or false;
   setBalloonScript = hypervisorConfig.setBalloonScript or null;
 
-  execArg = lib.optionalString microvmConfig.prettyProcnames
-    ''-a "microvm@${hostName}"'';
-
+  execArg = lib.optionalString microvmConfig.prettyProcnames ''-a "microvm@${hostName}"'';
 
   # TAP interface names for machined registration
   tapInterfaces = lib.filter (i: i.type == "tap" && i ? id) microvmConfig.interfaces;
   tapInterfaceNames = map (i: i.id) tapInterfaces;
+
+  # Generate machine UUID at eval time for consistency across SMBIOS and machined
+  # Uses provided machineId or generates UUIDv5 from hostname
+  machineUuid =
+    if microvmConfig.machineId != null then
+      microvmConfig.machineId
+    else
+      builtins.readFile (
+        vmHostPackages.runCommand "machine-uuid" { } ''
+          ${vmHostPackages.python3}/bin/python3 -c 'import uuid; print(uuid.uuid5(uuid.NAMESPACE_DNS, "${hostName}"), end="")' > $out
+        ''
+      );
 
   # Script to unregister from systemd-machined
   unregisterMachineScript = ''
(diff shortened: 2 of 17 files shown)


checks/machined.nix
```diff
@@ -0,0 +1,48 @@
+{ self, nixpkgs, system, hypervisor }:
+
+let
+  vmName = "machined-test";
+in
+{
+  # Test that MicroVMs can be registered with systemd-machined
+  "machined-${hypervisor}" = import (nixpkgs + "/nixos/tests/make-test-python.nix") ({ lib, ... }: {
```
SuperSandro2000: Should we not use pkgs.nixosTest?

lib/runner.nix
```diff
@@ -28,6 +28,88 @@ let
...
+  # TAP interface names for machined registration
+  tapInterfaces = lib.filter (i: i.type == "tap" && i ? id) microvmConfig.interfaces;
+  tapInterfaceNames = map (i: i.id) tapInterfaces;
+
+  # Generate machine UUID at eval time for consistency across SMBIOS and machined
+  # Uses provided machineId or generates UUIDv5 from hostname
+  machineUuid =
+    if microvmConfig.machineId != null
+    then microvmConfig.machineId
+    else builtins.readFile (vmHostPackages.runCommand "machine-uuid" {} ''
```
SuperSandro2000: We should try to avoid IFD as much as possible.

lib/runner.nix
```diff
@@ -28,6 +28,88 @@ let
...
+
+  # Script to register with systemd-machined
+  registerMachineScript = ''
+    set -euo pipefail
+
+    MACHINE_NAME="${hostName}"
+    UUID="${machineUuid}"
+
+    # Convert UUID to space-separated decimal bytes for busctl
+    UUID_BYTES=$(echo "$UUID" | tr -d '-' | ${vmHostPackages.gnused}/bin/sed 's/../0x& /g' | ${vmHostPackages.gawk}/bin/awk '{for(i=1;i<=NF;i++) printf "%d ", strtonum(''$i)}')
```
SuperSandro2000:
```suggestion
    UUID="${machineUuid}"

    PATH=${lib.makeBinPath (with vmHostPackages; [ gnused gawk systemd ])}

    # Convert UUID to space-separated decimal bytes for busctl
    UUID_BYTES=$(echo "$UUID" | tr -d '-' | sed 's/../0x& /g' | awk '{for(i=1;i<=NF;i++) printf "%d ", strtonum(''$i)}')
```

lib/runner.nix
```diff
@@ -28,6 +28,88 @@ let
...
+
+    ${if tapInterfaceNames == [] then ''
+    # No TAP interfaces, use simple RegisterMachine
+    ${vmHostPackages.systemd}/bin/busctl call \
```
SuperSandro2000:
```suggestion
    '' + if tapInterfaceNames == [] then ''
    # No TAP interfaces, use simple RegisterMachine
    busctl call \
```

lib/runner.nix
```diff
@@ -28,6 +28,88 @@ let
...
+    '' else ''
+    # Build network interface array for RegisterMachineWithNetwork
+    # Format: array of (ifindex, array of IP addresses)
+    IFINDICES=""
+    ${lib.concatMapStrings (name: ''
+    if [ -e /sys/class/net/${name}/ifindex ]; then
+      IFINDEX=$(cat /sys/class/net/${name}/ifindex)
+      IFINDICES="$IFINDICES $IFINDEX 0"
+    fi
+    '') tapInterfaceNames}
```
SuperSandro2000, line 78:
```suggestion
    '') tapInterfaceNames + ''
```

lib/runner.nix
```diff
@@ -28,6 +28,88 @@ let
...
+
+    # Count interfaces (pairs of ifindex + address count)
+    NUM_IFS=$(echo $IFINDICES | wc -w)
+    NUM_IFS=$((NUM_IFS / 2))
+
+    if [ "$NUM_IFS" -gt 0 ]; then
+      # Use RegisterMachineWithNetwork with TAP interfaces
+      ${vmHostPackages.systemd}/bin/busctl call \
```
SuperSandro2000, line 142:
```suggestion
      busctl call \
```

lib/runner.nix
```diff
@@ -28,6 +28,88 @@ let
...
+        org.freedesktop.machine1 \
+        /org/freedesktop/machine1 \
+        org.freedesktop.machine1.Manager \
+        RegisterMachineWithNetwork "sayssusa(iay)" \
+        "$MACHINE_NAME" \
+        16 $UUID_BYTES \
+        "microvm.nix" \
+        "vm" \
```
SuperSandro2000, line 150: It would probably be easier to read, if we don't split every word onto a new line.

lib/runner.nix
```diff
@@ -28,6 +28,88 @@ let
...
+        $$ \
+        "/" \
+        $NUM_IFS $IFINDICES
+    else
+      # Fallback to simple RegisterMachine
+      ${vmHostPackages.systemd}/bin/busctl call \
```
SuperSandro2000, line 155:
```suggestion
      busctl call \
```

lib/runners/qemu.nix
```diff
@@ -56,6 +56,14 @@ let
   inherit (microvmConfig) hostName vcpu mem balloon initialBalloonMem deflateOnOOM hotplugMem hotpluggedMem user interfaces shares socket forwardPorts devices vsock graphics storeOnDisk kernel initrdPath storeDisk credentialFiles;
   inherit (microvmConfig.qemu) machine extraArgs serialConsole pcieRootPorts;
 
+  # Generate machine UUID for SMBIOS - same logic as runner.nix
+  machineUuid =
+    if microvmConfig.machineId != null
+    then microvmConfig.machineId
+    else builtins.readFile (vmHostPackages.runCommand "machine-uuid" {} ''
```
SuperSandro2000: This is also IFD.

nixos-modules/microvm/default.nix
```diff
@@ -35,5 +35,12 @@ in
         inherit (config.system.build) toplevel;
       }
     );
+
+    # Set /etc/machine-id from machineId if provided
+    # This ensures the guest machine-id matches the UUID passed to machined and SMBIOS
+    environment.etc."machine-id" = lib.mkIf (config.microvm.machineId != null) {
+      mode = "0444";
+      text = builtins.replaceStrings ["-"] [""] config.microvm.machineId + "\n";
```
SuperSandro2000:
```suggestion
      text = lib.replaceString "-" "" config.microvm.machineId + "\n";
```

nixos-modules/microvm/default.nix
```diff
@@ -35,5 +35,12 @@ in
         inherit (config.system.build) toplevel;
       }
     );
+
+    # Set /etc/machine-id from machineId if provided
+    # This ensures the guest machine-id matches the UUID passed to machined and SMBIOS
+    environment.etc."machine-id" = lib.mkIf (config.microvm.machineId != null) {
+      mode = "0444";
```
SuperSandro2000:
```suggestion
```

nixos-modules/microvm/options.nix
```diff
@@ -462,6 +462,37 @@ in
...
+      description = ''
+        Register this MicroVM with systemd-machined on the host,
+        enabling management via machinectl.
+
+        When enabled, a registration script is generated in the
+        runner package. The host module will call this script
+        after the hypervisor starts.
+
+        The VM is registered with class "vm" using the UUID from
+        `machineId` (or a deterministic UUID derived from hostname).
```
SuperSandro2000: For easier readability we can make lines longer as 80 characters.

lib/runner.nix
```diff
@@ -41,6 +41,20 @@ let
...
+    set -euo pipefail
+    MACHINE_NAME="${hostName}"
+
+    # Terminate the machine registration (ignore errors if already gone)
+    ${vmHostPackages.systemd}/bin/busctl call \
+      org.freedesktop.machine1 \
+      /org/freedesktop/machine1 \
+      org.freedesktop.machine1.Manager \
+      TerminateMachine "s" \
+      "$MACHINE_NAME" 2>/dev/null || true
```
SuperSandro2000, line 56:
```suggestion
      "$MACHINE_NAME" 2>/dev/null
```
We already ignore the exit code in the systemd unit

lib/runner.nix
```diff
@@ -65,27 +79,24 @@ let
...
     '' else ''
-    # Build network interface array for RegisterMachineWithNetwork
-    # Format: array of (ifindex, array of IP addresses)
+    # Build network interface index array for RegisterMachineWithNetwork
     IFINDICES=""
     ${lib.concatMapStrings (name: ''
     if [ -e /sys/class/net/${name}/ifindex ]; then
-      IFINDEX=$(cat /sys/class/net/${name}/ifindex)
-      IFINDICES="$IFINDICES $IFINDEX 0"
+      IFINDICES="$IFINDICES $(cat /sys/class/net/${name}/ifindex)"
```
SuperSandro2000: Lets do this changes in the first commit touching those lines.

lib/runner.nix
```diff
@@ -79,24 +79,27 @@ let
...
     '' else ''
-    # Build network interface index array for RegisterMachineWithNetwork
+    # Build network interface array for RegisterMachineWithNetwork
+    # Format: array of (ifindex, array of IP addresses)
     IFINDICES=""
     ${lib.concatMapStrings (name: ''
     if [ -e /sys/class/net/${name}/ifindex ]; then
-      IFINDICES="$IFINDICES $(cat /sys/class/net/${name}/ifindex)"
+      IFINDEX=$(cat /sys/class/net/${name}/ifindex)
+      IFINDICES="$IFINDICES $IFINDEX 0"
```
SuperSandro2000: Why do we change those lines in `` fix: correct D-Bus signature for RegisterMachineWithNetwork`` again? Please try to keep a clean history even if it is less commits.

lib/runner.nix
SuperSandro2000, line 1: Please collapse the revert into the other commits.

nixos-modules/microvm/vsock-ssh.nix
```diff
@@ -0,0 +1,42 @@
...
+
+      From the host, connect using:
+      - For qemu/crosvm/kvmtool: `ssh vsock/<CID>`
+      - For cloud-hypervisor: `ssh vsock-mux/<path-to-notify.vsock>`
+
+      Or use: `microvm -s <vmname>`
+    '';
+  };
+
+  config = lib.mkIf (cfg.ssh.enable && cfg.cid != null) {
```
SuperSandro2000:
```suggestion
  config = lib.mkIf cfg.ssh.enable {
```
Otherwise, the assertion below does never trigger.

nixos-modules/microvm/vsock-ssh.nix
```diff
@@ -0,0 +1,42 @@
...
+    assertions = [{
+      assertion = cfg.cid != null;
+      message = "microvm.vsock.ssh.enable requires microvm.vsock.cid to be set";
+    }];
+
+    services.openssh = {
+      enable = true;
+      startWhenNeeded = true;
```
SuperSandro2000: Lets not change that for most people and keep the ssh defaults.

nixos-modules/microvm/vsock-ssh.nix
```diff
@@ -0,0 +1,42 @@
...
+      # Don't listen on network interfaces by default when using VSOCK
+      # Users can override this if they want both
+      listenAddresses = lib.mkDefault [];
```
SuperSandro2000: This would break the microvms of most users I know. Speaks there anything against keeping the defaults?

pkgs/microvm-command.nix
```diff
@@ -195,5 +201,52 @@ writeShellScriptBin "microvm" ''
...
+          SSH_TARGET="vsock-mux/$DIR/notify.vsock"
+          ;;
+        *)
+          # qemu, crosvm, kvmtool use native AF_VSOCK with CID
+          SSH_TARGET="vsock/$VSOCK_CID"
+          ;;
+      esac
+
+      echo -e "${colored "boldCyan" "Connecting to $NAME via VSOCK..."}"
+      exec ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null "$SSH_TARGET" "$@"
```
SuperSandro2000: Why do we disable StrictHostKeyChecking and the UserKnownHostsFile?

lib/runner.nix
SuperSandro2000, line 1: `fix: register correct leader PID with machined` should be squashed into the original commit adding the feature.

lib/runners/qemu.nix
SuperSandro2000, line 1: Same for ` fix: handle missing control socket in qemu shutdown script `

nixos-modules/microvm/options.nix
SuperSandro2000, line 1: same for ` docs: document machinectl limitations for VMs `

nixos-modules/microvm/default.nix
elsbrock: Hmm, that doesn't exist in my `nix`

pkgs/microvm-command.nix
elsbrock: commented above

nixos-modules/microvm/default.nix
SuperSandro2000: I think it is new with 25.11.

pkgs/microvm-command.nix
SuperSandro2000: Ah, I think that makes sense then to disable it.

nixos-modules/microvm/options.nix
```diff
@@ -462,6 +462,40 @@ in
...
+        - `terminate`, `kill` - stop VM (will auto-restart if Restart=always)
+
+        Note: `machinectl reboot` stops the VM but won't auto-restart it because systemd treats it as an
+        intentional stop. Use `systemctl restart microvm@<name>` for restarts.
+      '';
+    };
+
+    machineId = mkOption {
+      type = with types; nullOr str;
+      default = null;
```
astro: Why not put in the `hostName`-based generation as default here, instead of doing the extra `makeMachineUuid` logic?

nixos-modules/microvm/options.nix
```diff
@@ -462,6 +462,40 @@ in
       '';
     };
 
+    registerWithMachined = mkOption {
+      type = types.bool;
+      default = false;
```
astro, line 491: There are no downsides listed in the description. Are there reasons to not enable this cool feature by default?

elsbrock: None that I'm aware of at least. Maybe somebody knows?

nixos-modules/host/default.nix
```diff
@@ -113,6 +113,19 @@ in
...
+        ];
+        after = lib.optionals runner.registerWithMachined [
+          "systemd-machined.service"
+        ];
+        serviceConfig.ExecStartPost = lib.optionals runner.registerWithMachined [
+          "+${stateDir}/${name}/current/bin/microvm-register $MAINPID"
+        ];
+        serviceConfig.ExecStopPost = lib.optionals runner.registerWithMachined [
+          "+${stateDir}/${name}/current/bin/microvm-unregister"
+        ];
```
astro: This covers only the declarative case where a discete `microvm@${hostName}` unit is generated.

We also want to support imperative mode with the `microvm@` template unit. Can we somehow call `microvm-{un,}register` at runtime, without systemd panicking if it doesn't exist when the feature is disabled?

The easiest solution would be to call scripts that test whether `microvm-{,un}register` exist.

nixos-modules/microvm/options.nix
astro: I did that in my branch until this PR gets merged.

pkgs/microvm-command.nix
```diff
@@ -195,5 +203,54 @@ writeShellScriptBin "microvm" ''
...
+        *)
+          # qemu, crosvm, kvmtool use native AF_VSOCK with CID
+          SSH_TARGET="vsock/$VSOCK_CID"
+          ;;
+      esac
+
+      echo -e "${colored "boldCyan" "Connecting to $NAME via VSOCK..."}"
+      # VSOCK is a local host-to-guest transport without network exposure,
+      # and VM host keys change on each rebuild - skip known_hosts checks.
+      exec ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null "$SSH_TARGET" "$@"
```
astro: I can run this as a non-root user, so it might make sense to specify `root@`:

```suggestion
      exec ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -l root "$SSH_TARGET" "$@"
```

nixos-modules/microvm/options.nix
SuperSandro2000: `machinectl reboot` does not work OOTB.

elsbrock: Yea, let's add that. More of a limitation though.

pkgs/microvm-command.nix
SuperSandro2000: @elsbrock wdyt?

nixos-modules/microvm/system.nix
```diff
@@ -58,5 +58,14 @@
         generators = { systemd-gpt-auto-generator = "/dev/null"; };
       };
 
+    # Set /etc/machine-id from machineId if provided
+    # This ensures the guest machine-id matches the UUID passed to machined and SMBIOS
+    environment.etc."machine-id" = lib.mkIf (config.microvm.machineId != null) {
+      text = builtins.replaceStrings ["-"] [""] config.microvm.machineId + "\n";
```
SuperSandro2000:
```suggestion
      text = lib.replaceString "-" "" config.microvm.machineId + "\n";
```

nixos-modules/microvm/options.nix
```diff
@@ -462,6 +462,51 @@ in
...
+
+        Note: `machinectl reboot` stops the VM but won't auto-restart it because systemd treats it as an
+        intentional stop. Use `systemctl restart microvm@<name>` for restarts.
+      '';
+    };
+
+    machineId = mkOption {
+      type = with types; nullOr str;
+      default =
+        let
```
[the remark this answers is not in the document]
SuperSandro2000: Other runners do not use it.

nixos-modules/host/default.nix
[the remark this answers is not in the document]
SuperSandro2000: Out of scope

pkgs/microvm-command.nix
```diff
@@ -195,5 +206,61 @@ writeShellScriptBin "microvm" ''
...
+      SSH_REMOTE_CMD=()
+      if [ "$#" -gt 0 ] && [ "$1" = "--" ]; then
+        shift
+        SSH_EXTRA_OPTS=("$@")
+      else
+        SSH_REMOTE_CMD=("$@")
+      fi
+
+      echo -e "${colored "boldCyan" "Connecting to $NAME via VSOCK..."}"
+      exec ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -l root "''${SSH_EXTRA_OPTS[@]}" "$SSH_TARGET" "''${SSH_REMOTE_CMD[@]}"
```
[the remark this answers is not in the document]
SuperSandro2000: That is by design by systemd upstream.

checks/machined.nix
```diff
@@ -0,0 +1,77 @@
...
+            # Verify machine status works
+            host.succeed("machinectl status '${vmName}'")
+
+            # Verify the machine class is 'vm'
+            host.succeed("machinectl show '${vmName}' --property=Class | grep -q 'vm'")
+
+            ${lib.optionalString (hypervisor != "cloud-hypervisor") ''
+              # On systemd >=259 RegisterMachineEx path should expose VSOCK/SSH metadata
+              host.succeed("machinectl show '${vmName}' --property=VSockCID | grep -q 'VSockCID=${toString vsockCid}'")
+              host.succeed("machinectl show '${vmName}' --property=SSHAddress | grep -q 'SSHAddress=vsock/${toString vsockCid}'")
```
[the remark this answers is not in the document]
SuperSandro2000: Haha, not a concern for NixOS

SuperSandro2000 (approved):

nixos-modules/microvm/system.nix
```diff
@@ -58,5 +58,14 @@
         generators = { systemd-gpt-auto-generator = "/dev/null"; };
       };
 
+    # Set /etc/machine-id from machineId if provided
+    # This ensures the guest machine-id matches the UUID passed to machined and SMBIOS
+    environment.etc."machine-id" = lib.mkIf (config.microvm.machineId != null) {
+      text = lib.replaceString "-" "" config.microvm.machineId + "\n";
+    };
+    # Generate hostId from machine-id like systemd would do
+    networking.hostId = lib.mkDefault (
```
oddlama, line 67: if a user sets `machineId = null` this will error out. should probably be guarded with mkIf machineId != null

nixos-modules/microvm/options.nix
```diff
@@ -486,6 +486,53 @@ in
...
+        - `terminate`, `kill` - stop VM (will auto-restart if Restart=always)
+
+        Note: `machinectl reboot` stops the VM but won't auto-restart it because systemd treats it as an
+        intentional stop. Use `systemctl restart microvm@<name>` for restarts.
+      '';
+    };
+
+    machineId = mkOption {
+      type = with types; nullOr str;
+      default =
```
oddlama, line 510: This default bit me when updating, as it overwrote the generated machine-id that was in use previously (which caused a lot of dependent failures). This change is a breaking change for existing declarative instances

nixos-modules/microvm/system.nix
SuperSandro2000: I've opened #499, but I only eye balled this, so it could be another issue like this in the code base.

nixos-modules/microvm/options.nix
SuperSandro2000: I do not have a good idea how to make this migration smooth unless we introduce a state version.

PR #456, opened 2026-01-08
qemu: enable rtc

SuperSandro2000:
Before we had no /dev/rtc
Sadly the qemu microvm docs are not clear about the default.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -5,7 +5,6 @@ extend-exclude = [
 ]
 
 [default.extend-words]
-
+substituters = "substituters"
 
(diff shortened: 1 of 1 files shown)


.typos.toml
```diff
@@ -5,7 +5,6 @@ extend-exclude = [
 ]
 
 [default.extend-words]
-
+substituters = "substituters"
```
astro, line 8: wat?

SuperSandro2000: typos maps a typo word to the corrected result or if you want to say that word is valid: map it to itself.

PR #471, opened 2026-02-17
Examples refactor/improvement

randomizedcoder:
G'day mate,

I've been really using microvm-s a lot for testing, and they work great!

I found that I couldn't run ebpf programs inside the microvms, so added a BTF option, and added an example /examples/btf-vhosts to demonstrate BTF and vhost-net works.

I also realized that the serial port and high speed console didn't have a great example, so I started adding that, which lead to refactoring the examples folder.

I've added detailed documentation about the examples, added more examples, and added automated tests.

```
nix run .#test-all-examples
```

It actually looks like the reason they are slow slow to boot is initrd, which maybe you knew about already.

Hopefully this all makes sense.

Thanks,
Dave

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,6 +17,10 @@ nix run .#cloud-hypervisor-example
 # Run all checks (comprehensive test matrix)
 nix flake check
 
+# Run automated example tests (see examples/README.md for details)
+nix run .#test-all-examples           # Run all example tests
+nix run .#test-all-examples-repeat    # Run 3 times (catch flaky failures)
+
 # Build documentation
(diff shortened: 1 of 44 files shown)


CLAUDE.md
SuperSandro2000, line 1: Claude is the whole reason why I started to engineer this together as they are to greedy to build rate limiting or any kind of back off into their never ending scraping. https://gitea.c3d2.de/c3d2/nix-config/src/commit/943ab67be3cb3f040453ab248b4f085a6fdcf370/modules/nginx.nix#L164-L199

So no.

Issue #474, opened 2026-02-22
notify slowing down cloud-hypervisor boot


yorickvP:
When I set `microvm.vsock.cid = 4; hypervisor = "cloud-hypervisor";` and run my microvm as a systemd unit, the VM takes about 30 seconds to boot. When I run it without the cid, it takes 6 seconds to boot.

PR #493, opened 2026-03-27
Fix 2-second systemd notify stalls in cloud-hypervisor

rzhikharevich:
This is a fix for #474. See the code comment for explanation and caveats.

Relevant parts of:
- systemd: [sd-daemon.c:630](https://github.com/systemd/systemd/blob/5f2258affcbcf7690da9655308a1f308d0baf7fd/src/libsystemd/sd-daemon/sd-daemon.c#L630)
- [cloud-hypervisor](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/c06f6c32932e3241676ba658a8fa9b2852d9e305/virtio-devices/src/vsock/csm/connection.rs#L339)

### DIFF
--- a/lib/runners/cloud-hypervisor.nix
+++ b/lib/runners/cloud-hypervisor.nix
@@ -138,27 +138,28 @@ let
   # An important caveat is that the message could theoretically be truncated but this should not
   # happen in practice unless the message is large (systemd notifications are small).
   #
-  # TODO(rzhikharevich): Ideally, cloud-hypervisor should propagate the half-close, then the
+  # TODO: Ideally, cloud-hypervisor should propagate the half-close, then the
   # theoretical truncation risk could be fixed by closing the connection after detecting it.
   vsockNotifyProxy = pkgs.writers.writePython3Bin "vsock-notify-proxy" {} ''
     import asyncio
     import os
     import socket
+    import sys
 
-    notify_socket = os.environ["NOTIFY_SOCKET"]
     notify = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+    notify.connect(os.environ["NOTIFY_SOCKET"])
 
 
     async def handle(reader, writer):
         data = await reader.read(65536)
         writer.close()
         if data:
-            notify.sendto(data, notify_socket)
+            notify.send(data)
+        await writer.wait_closed()
 
 
     async def main():
-        path = os.environ["VSOCK_NOTIFY_PATH"]
-        server = await asyncio.start_unix_server(handle, path=path)
+        server = await asyncio.start_unix_server(handle, path=sys.argv[1])
         await server.serve_forever()
 
 
@@ -196,7 +197,7 @@ in {
 
     # Forward systemd notify messages from guest (via vsock) to host
     if [ -n "''${NOTIFY_SOCKET:-}" ]; then
-      VSOCK_NOTIFY_PATH=${vsockPath}_8888 ${vsockNotifyProxy}/bin/vsock-notify-proxy &
+      ${vsockNotifyProxy}/bin/vsock-notify-proxy ${vsockPath}_8888 &
     fi
   '' + lib.optionalString graphics.enable ''
     rm -f ${graphics.socket}
--- a/lib/runners/cloud-hypervisor.nix
+++ b/lib/runners/cloud-hypervisor.nix
@@ -126,6 +126,45 @@ let
     vulkan = true;
   };
 
+  # A lightweight proxy that forwards systemd notifications from a vsock
+  # Unix socket to the host's NOTIFY_SOCKET. Unlike socat, it closes the
+  # connection immediately after reading, which unblocks the guest's
+  # blocking recv() (systemd does shutdown(SHUT_WR) + recv() as an
+  # end-of-message handshake, but cloud-hypervisor does not propagate the
+  # half-close to the host-side Unix socket).
+  vsockNotifyProxy = pkgs.writers.writePython3Bin "vsock-notify-proxy" {} ''
+    import asyncio
+    import os
+    import socket
+
+    notify_socket = os.environ["NOTIFY_SOCKET"]
+    notify = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+
+
+    async def handle(reader, writer):
+        chunks = []
+        while True:
+            chunk = await reader.read(65536)
+            if not chunk:
+                break
+            chunks.append(chunk)
+            if chunk.endswith(b"\n"):
+                break
+        writer.close()
+        data = b"".join(chunks)
+        if data:
+            notify.sendto(data, notify_socket)
+
+
+    async def main():
+        path = os.environ["VSOCK_NOTIFY_PATH"]
+        server = await asyncio.start_unix_server(handle, path=path)
+        await server.serve_forever()
+
+
+    asyncio.run(main())
+  '';
+
   oemStringValues = platformOEMStrings ++ lib.optional supportsNotifySocket "io.systemd.credential:vmm.notify_socket=vsock-stream:2:8888";
   oemStringOptions = lib.optional (oemStringValues != []) "oem_strings=[${lib.concatStringsSep "," oemStringValues}]";
   platformExtracted = extractOptValues "--platform" extraArgs;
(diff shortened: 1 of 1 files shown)


lib/runners/cloud-hypervisor.nix
```diff
@@ -155,11 +194,9 @@ in {
     # Ensure notify sockets are removed if cloud-hypervisor didn't exit cleanly the last time
     rm -f ${vsockPath} ${vsockPath}_8888
 
-    # Start socat to forward systemd notify socket over vsock
+    # Forward systemd notify messages from guest (via vsock) to host
     if [ -n "''${NOTIFY_SOCKET:-}" ]; then
-      # -T2 is required because cloud-hypervisor does not handle partial
-      # shutdown of the stream, like systemd v256+ does.
-      ${pkgs.socat}/bin/socat -T2 UNIX-LISTEN:${vsockPath}_8888,fork UNIX-SENDTO:$NOTIFY_SOCKET &
+      VSOCK_NOTIFY_PATH=${vsockPath}_8888 ${vsockNotifyProxy}/bin/vsock-notify-proxy &
```
```suggestion
      ${lib.getExe vsockNotifyProxy} ${vsockPath}_8888 &
```

lib/runners/cloud-hypervisor.nix
```diff
@@ -126,6 +126,45 @@ let
...
+    async def handle(reader, writer):
+        data = await reader.read(65536)
+        writer.close()
+        if data:
+            notify.sendto(data, notify_socket)
+
+
+    async def main():
+        path = os.environ["VSOCK_NOTIFY_PATH"]
+        server = await asyncio.start_unix_server(handle, path=path)
```
```suggestion
        server = await asyncio.start_unix_server(handle, path=sys.argv[0])
```
We can make this kinda dirty here, as we control all invocations and this fails loudly when no arg is there, os.environ does IIRC return empty string.

lib/runners/cloud-hypervisor.nix
```diff
@@ -126,6 +126,45 @@ let
...
+  #
+  # An important caveat is that the message could theoretically be truncated but this should not
+  # happen in practice unless the message is large (systemd notifications are small).
+  #
+  # TODO(rzhikharevich): Ideally, cloud-hypervisor should propagate the half-close, then the
+  # theoretical truncation risk could be fixed by closing the connection after detecting it.
+  vsockNotifyProxy = pkgs.writers.writePython3Bin "vsock-notify-proxy" {} ''
+    import asyncio
+    import os
+    import socket
```
SuperSandro2000, line 151: glad we do not need anything outside the stdenv.

lib/runners/cloud-hypervisor.nix
```diff
@@ -126,6 +126,45 @@ let
...
+  # end-of-message handshake in sd-daemon.c, but cloud-hypervisor does not propagate the half-close
+  # to the host-side Unix socket).
+  #
+  # Previously, this used socat with a 2-second timeout but this causes systemd to stall on every
+  # notification, causing extremely slow boot times.
+  #
+  # An important caveat is that the message could theoretically be truncated but this should not
+  # happen in practice unless the message is large (systemd notifications are small).
+  #
+  # TODO(rzhikharevich): Ideally, cloud-hypervisor should propagate the half-close, then the
```
```suggestion
  # TODO: Ideally, cloud-hypervisor should propagate the half-close, then the
```

lib/runners/cloud-hypervisor.nix
```diff
@@ -126,6 +126,45 @@ let
...
+    import asyncio
+    import os
+    import socket
+
+    notify_socket = os.environ["NOTIFY_SOCKET"]
+    notify = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+
+
+    async def handle(reader, writer):
+        data = await reader.read(65536)
```
SuperSandro2000, line 159: Can we detect with a one line if we read to little and print that to stdout?

lib/runners/cloud-hypervisor.nix
```diff
@@ -126,6 +126,45 @@ let
...
+        writer.close()
+        if data:
+            notify.sendto(data, notify_socket)
+
+
```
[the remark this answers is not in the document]
SuperSandro2000: I do not do async python programming, so I do not know if that is a good solution or not.

lib/runners/cloud-hypervisor.nix
```diff
@@ -126,6 +126,45 @@ let
...
+    notify_socket = os.environ["NOTIFY_SOCKET"]
+    notify = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+
+
+    async def handle(reader, writer):
+        data = await reader.read(65536)
+        writer.close()
+        if data:
+            notify.sendto(data, notify_socket)
```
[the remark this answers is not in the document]
SuperSandro2000: I would also be fine to just error out if notify_socket starts with `@`.

RaitoBezarius:
> @RaitoBezarius Please help review.

I'm sorry, I don't review anything that involves GenAI in a project on my free time.

lib/runners/cloud-hypervisor.nix
```diff
@@ -126,6 +126,45 @@ let
...
+    import os
+    import socket
+
+    notify_socket = os.environ["NOTIFY_SOCKET"]
+    notify = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+
+
+    async def handle(reader, writer):
+        data = await reader.read(65536)
+        writer.close()
```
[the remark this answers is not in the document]
astro: The 1st example in the [Python asyncio documentation](https://docs.python.org/3/library/asyncio-stream.html) does this, too.

lib/runners/cloud-hypervisor.nix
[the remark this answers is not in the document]
astro: `reader.read(-1)` will always read until EOF according to the docs. Isnt that what we want instead?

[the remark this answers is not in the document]
yorickvP: Apparently, cloud-hypervisor doesn't pass the EOF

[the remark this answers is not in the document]
SuperSandro2000: Hmmmmm, so we just hope the 64k is enough. Would be fine for me if the average message is way below that, which I think it should be.

SuperSandro2000:
@rzhikharevich can you please incorporate the feedback?

lib/runners/cloud-hypervisor.nix
rzhikharevich: Updated as suggested

lib/runners/cloud-hypervisor.nix
rzhikharevich: Updated (n.b. `sys.argv[1]`).

FYI `os.environ` does throw an exception if the variable is unset.

lib/runners/cloud-hypervisor.nix
rzhikharevich: Updated

lib/runners/cloud-hypervisor.nix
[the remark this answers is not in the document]
rzhikharevich: Done

rzhikharevich:
@SuperSandro2000 Hey, sorry for the delay, updated the PR

lib/runners/cloud-hypervisor.nix
```diff
@@ -130,6 +130,46 @@ let
     vulkan = true;
   };
 
+  # A proxy that forwards systemd notifications from cloud-hypervisor's vsock socket. It
+  # intentionally calls read only once and closes the connection immediately, which unblocks the
+  # guest's blocking recv() (systemd does shutdown(SHUT_WR) + recv() as an
+  # end-of-message handshake in sd-daemon.c, but cloud-hypervisor does not propagate the half-close
+  # to the host-side Unix socket).
+  #
+  # Previously, this used socat with a 2-second timeout but this causes systemd to stall on every
```
astro, line 140: If that's a meaningful change that users should know about, then please document it in `/CHANGELOG.md`

lib/runners/cloud-hypervisor.nix
astro: This shouldn't be a `TODO` for microvm.nix, we're already done by adding such workarounds for all affected versions. I think we should link to an upstream ticket in cloud-hypervisor here. :)

rzhikharevich:
cloud-hypervisor v53 now propagates half-closed sockets + commit 9a67f3f is a sensible workaround so I'm closing this PR.

PR #495, opened 2026-03-30
feat: add storeOnDisk support for vfkit runner

luizribeiro:
Passes the store disk image as a read-only `virtio-blk` device when `storeOnDisk = true`, consistent with other runners.

This avoids sharing `/nix/store` from the macOS host via virtiofs, which is problematic due to macOS's case-insensitive filesystem. See comments in #154.

### DIFF
--- a/lib/runners/vfkit.nix
+++ b/lib/runners/vfkit.nix
@@ -13,7 +13,7 @@ let
 
   inherit (microvmConfig)
     vcpu mem user interfaces shares socket
-    storeOnDisk kernel initrdPath kernelParams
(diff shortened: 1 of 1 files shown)


bpinto:
I'm not sure if any action is needed, however bubblewrap won't work on vfkit (macOS), so this [fallback](https://github.com/microvm-nix/microvm.nix/blob/0f112113172c9fabeb27a94f39397d11df49e671/nixos-modules/microvm/store-disk.nix#L91) is always triggered.

> Bubblewrap failed. Falling back to copying...

We may continue to ignore this, or we could create an alternative path that doesn't use bubblewrap?

PR #498, opened 2026-04-02
fix(firecracker): remove noapic acpi=off from Firecracker boot_args for PCI mode

t4min0:
PR #428 added `--enable-pci` for Firecracker >= 1.13.0, but the hardcoded `boot_args` still contain `noapic acpi=off`. With `acpi=off`, the guest kernel cannot parse the ACPI MCFG table Firecracker provides to discover the ECAM region, so PCI device enumeration fails and the VM hangs at boot.

`noapic` disables the IO-APIC, which is unnecessary with PCI transport since virtio PCI devices use MSI-X (bypasses IO-APIC entirely).

Removing both params resolves the boot hang and allows `--enable-pci` to work as intended.

Tested on a 14-VM Firecracker fleet (NixOS, kernel 6.12, Firecracker 1.13.2): all VMs boot and run correctly with PCI transport after this change.

Fixes #496

### DIFF
--- a/lib/runners/firecracker.nix
+++ b/lib/runners/firecracker.nix
@@ -24,7 +24,7 @@ let
     boot-source = {
       kernel_image_path = kernelPath;
       initrd_path = initrdPath;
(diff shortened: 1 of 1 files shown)


astro:
Thank you!

PR #503, opened 2026-04-12
Explicitly set user in supervisord config

ncaq:
The `microvm-virtiofsd@` service runs as root by design,
because virtiofsd needs root privileges to set `--rlimit-nofile 1048576`.
However, the supervisord configuration did not explicitly declare `user = "root"`,
causing supervisord to emit a CRIT-level warning on every start:

CRIT Supervisor is running as root.
Privileges were not dropped because no user is specified in the config file.
If you intend to run as root,
you can set user=root in the config file to avoid this message.

This commit adds `user = "root"` to the `[supervisord]` section of the generated config to suppress this warning,
making it explicit that running as root is intentional.

### DIFF
--- a/nixos-modules/microvm/virtiofsd/default.nix
+++ b/nixos-modules/microvm/virtiofsd/default.nix
@@ -16,7 +16,10 @@ in
     virtiofsd-run =
       let
         supervisordConfig = {
-          supervisord.nodaemon = true;
+          supervisord = {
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
We don't reactor anything here, we fix something, but since we don't auto generate something from commit messages it is rather pointless to add random prefixes.

PR #507, opened 2026-04-20
feat: create share paths on host

Liassica:
#316 seems to be more or less abandoned; this implements effectively the same change but using systemd-tmpfiles as suggested in https://github.com/microvm-nix/microvm.nix/pull/316#discussion_r3016587127.


astro:
Thank you!

Issue #508, opened 2026-04-22
nixos-rebuild switch does not stop orphaned microvm@<name>.service instances when a VM is removed from config


brandonros:
## Summary

When a VM is declared via `microvm.vms.<name>` (with `microvm.autostart` including it, or after a manual `systemctl start microvm@<name>`) and is later removed from the NixOS config, `nixos-rebuild switch` does not stop the running `microvm@<name>.service` instance. The qemu process keeps running, `Restart=always` brings it back if killed, and `/var/lib/microvms/<name>/` is never cleaned up. This breaks the mental model of NixOS as a reconciling system, since removing a declaration silently leaves the old VM running indefinitely.

## Repro

1. Declare two VMs, e.g. `microvm.vms.foo` and `microvm.vms.bar`, and add both to `microvm.autostart`.
2. `nixos-rebuild switch` — both VMs start.
3. Comment out `microvm.vms.bar` in config.
4. `nixos-rebuild switch`.
5. `systemctl is-active microvm@bar.service` still reports `active`. The qemu process is still running. `/var/lib/microvms/bar/` and its disk image are still on disk.

## Expected

Removing a VM from `microvm.vms` and running `nixos-rebuild switch` should at minimum **stop the running `microvm@<name>.service` instance** (matching the convention for ordinary `systemd.services.<name>`, whose units are stopped when removed from config). Disk state under `/var/lib/microvms/<name>/` can reasonably stay (matches how other services leave state in `/var/lib/`), but the qemu process should not survive its declaration being removed.

## Root cause

Traced through `nixos-modules/host/default.nix` at commit HEAD:

**Positive case (VM declared) is handled:**
- [`host/default.nix` L64-L285](https://github.com/microvm-nix/microvm.nix/blob/main/nixos-modules/host/default.nix#L64-L285) — `foldl'` over `attrNames config.microvm.vms` generates per-VM units (`install-microvm-<name>`, `microvm@<name>`, etc.).
- [`host/default.nix` L287-L293](https://github.com/microvm-nix/microvm.nix/blob/main/nixos-modules/host/default.nix#L287-L293) — `microvms.target.wants` is rebuilt from scratch each activation, pointing only at currently-declared VMs.

**Negative case (VM removed) is not handled:**
[...]

Issue #302, opened 2024-12-02
Deleting microvms does not delete all associated resources


pecigonzalo:
Hi! Im declaring VMs with the declarative approach inside of my NixOS config. When microvms are deleted, there are some network resources that seem to be left behind and not removed.

- tuntap devices
- microvms themeselves in `/var/lib/microvms`

Shouldn't these be cleaned up when the microvm is undeclared?

Thanks

romner-set:
Also running into this after upgrading to 24.11, it's causing my VMs to completely lose networking after `microvm -Ru` or similar commands.

For me it's happening in the `microvm-tap-interfaces@` stop script, specifically [this line](https://github.com/astro/microvm.nix/blob/e8d5f12b834a59187c7ec147a8952a0567f97939/nixos-modules/host/default.nix#L143) is erroring out with `ioctl(TUNSETIFF): Invalid argument`. Replacing the command with `ip link del $id` seems to fix it.

astro:
Thanks!

It seems that `ip link delete` is much less picky than `ip tuntap del`, so let's use it!

pecigonzalo:
Chiming back in here, I don’t think it’s just the links. The VMs are left
running for example. I would expect this to work like the container NixOS
module that removes them if removed from the declared code.

Does that make sense?

astro:
The systemd units are explicitly not restarted on the host's `nixos-rebuild switch` because they are updated independently.

What do you think about adding a section to the FAQ in the handbook? Would that help?

astro:
The VMs are updated independently when managing them imperatively via `microvm -u`.

I'd accept PRs that stop declarative VMs on removal but I'm very vary of automatically removing state for VMs that are perhaps only deactivated temporarily.

pecigonzalo:
Yeah, in the imperative use case I can see it, but in the declarative they
are not behaving as expected right?

I’ll try and cook up a PR

pecigonzalo:
@astro I think the change would be to set [restartIfChanged](https://search.nixos.org/options?channel=24.11&show=systemd.services.%3Cname%3E.restartIfChanged&from=50&size=50&sort=relevance&type=packages&query=services.%3Cname) to true (by removing it). 

This is what I believe makes the difference in the `container-oci`, but its explicitly set to `false` in `microvm.nix` [eg.](https://github.com/astro/microvm.nix/blob/main/nixos-modules/host/default.nix#L134).
[...]

PR #509, opened 2026-04-22
host: opt-in stop for orphaned microvm@<name>.service on removal from microvm.vms

brandonros:
## Summary

Closes the reconciliation gap where `nixos-rebuild switch` does not stop a `microvm@<name>.service` after `<name>` is removed from `microvm.vms`. Addresses the stop-on-removal goal from #302 (@astro: *""I'd accept PRs that stop declarative VMs on removal""*), scoped to stopping only — state under `/var/lib/microvms/<name>/` is preserved per the same comment.

Tracked as #508.

**Opt-in by default.** New behavior is gated behind `microvm.host.stopOrphans` which defaults to `false` — existing users see no change. Users who want NixOS-style reconciliation for their declarative VMs set `microvm.host.stopOrphans = true;`.

## Why `restartIfChanged` / the prior #302 discussion doesn't cover this

The fix briefly discussed in #302 (flipping `restartIfChanged` to match `oci-containers`) never shipped, and would not have fixed this bug. `X-RestartIfChanged` controls **restart on config change**; stop-on-removal is a separate path in `switch-to-configuration-ng` (`collect_unit_changes`, `src/main.rs:1078-1089`) that checks whether the service's *base unit file* disappeared.

For `microvm@<name>.service`, the base unit file is the template `microvm@.service`, which the host module unconditionally defines at `nixos-modules/host/default.nix:240`. So `new_base_unit_file` always exists and the stop-on-removal branch never fires — regardless of `restartIfChanged`. The `.wants` symlinks under `microvms.target.wants/` are not examined either (no references in the Rust source).

`microvm.vms.<name>.restartIfChanged` already exists as a per-VM option (#104, #110); that half of the #302 agreement effectively landed. The stop-on-removal half is the remaining gap.

## What this PR does

Adds `system.activationScripts.microvm-stop-orphans` to `nixos-modules/host/default.nix`. When enabled, on each activation it:

1. Lists `install-microvm-*.service` files in `/run/current-system/etc/systemd/system/` — these are the VMs that were declarative in the previous generation.
2. For each, checks if its `<name>` is still in `attrNames config.microvm.vms` of the new generation.
3. If not, and `microvm@<name>.service` is currently active, calls `systemctl stop microvm@<name>.service`.

Using `install-microvm-<name>.service` as the declarative-VM marker is important: imperative VMs managed via the `microvm` command have no `install-microvm-<name>.service` unit, so they are never touched by this reconciliation.

## What this PR explicitly does NOT do

- Does **not** change default behavior — opt-in via `microvm.host.stopOrphans = true;`.
- Does **not** remove `/var/lib/microvms/<name>/` or any VM state, matching the concern in [#302](https://github.com/microvm-nix/microvm.nix/issues/302#issuecomment-2549214467). State cleanup remains manual, as documented in `doc/src/microvm-command.md`.
- Does **not** affect imperative MicroVMs, since they lack the `install-microvm-<name>.service` marker.
- Does **not** touch `Restart=always`, `restartIfChanged`, or any other lifecycle semantics.

## Files changed

- `nixos-modules/host/default.nix` — new activation script (16 lines, inert when the option is `false`).
- `nixos-modules/host/options.nix` — new `microvm.host.stopOrphans` option (default `false`).
- `doc/src/declarative.md` — new ""Reconciliation on host rebuild"" section, framed as opt-in.
- `doc/src/microvm-command.md` — split ""Removing MicroVMs"" into Imperative / Declarative / State cleanup subsections; declarative path mentions the opt-in.

## Test plan

I don't have a NixOS test host to iterate on, and local evaluation of the full flake checks exhausts my machine's memory, so this relies on CI. Happy to add a nixosTest in a follow-up — the natural shape is to declare two VMs with `microvm.host.stopOrphans = true;`, switch to a specialisation that drops one, and assert `microvm@<dropped>.service` is inactive while `/var/lib/microvms/<dropped>/` remains.

- [ ] CI evaluation passes
- [ ] Manual verification on a NixOS host with `microvm.host.stopOrphans = true;` set
- [ ] nixosTest added in follow-up (if desired)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/doc/src/declarative.md
+++ b/doc/src/declarative.md
@@ -71,3 +71,31 @@ microvm.vms = {
 
 Note that building MicroVMs with the host increases build time and
 closure size of the host's system.
+
+## Reconciliation on host rebuild
+
+By default, removing a VM from `microvm.vms.<name>` does **not** stop
+the running `microvm@<name>.service`. Out of the box, MicroVM
+lifecycle is decoupled from host configuration — updates are applied
+via the imperative [`microvm` command](./microvm-command.md), and
+removing a declaration simply stops managing the VM declaratively
+without disturbing the running process.
+
+To opt in to reconciliation — so that removing a VM from
+`microvm.vms.<name>` and running `nixos-rebuild switch` stops the
+corresponding `microvm@<name>.service`, matching the behavior of
+ordinary `systemd.services.<name>` — set:
+
+```nix
+microvm.host.stopOrphans = true;
+```
+
+When enabled:
+
+- State under `/var/lib/microvms/<name>/` is preserved, in case the VM
+  is only being temporarily deactivated. To fully remove a VM including
+  its disk images, `rm -rf /var/lib/microvms/<name>` after the service
+  has stopped.
+- Imperative MicroVMs managed via the `microvm` command are never
+  touched; only VMs that were previously declared in `microvm.vms` are
+  considered.
--- a/doc/src/declarative.md
+++ b/doc/src/declarative.md
@@ -74,17 +74,14 @@ closure size of the host's system.
 
 ## Reconciliation on host rebuild
 
-By default, removing a VM from `microvm.vms.<name>` does **not** stop
-the running `microvm@<name>.service`. Out of the box, MicroVM
-lifecycle is decoupled from host configuration — updates are applied
-via the imperative [`microvm` command](./microvm-command.md), and
-removing a declaration simply stops managing the VM declaratively
-without disturbing the running process.
+By default, removing a VM from `microvm.vms.<name>` does **not** stop the running `microvm@<name>.service`,
+because MicroVM lifecycle is decoupled from host configuration.
+Updates are meant to be applied via the imperative [`microvm` command](./microvm-command.md),
+and removing a declaration simply stops managing the VM declaratively.
 
-To opt in to reconciliation — so that removing a VM from
-`microvm.vms.<name>` and running `nixos-rebuild switch` stops the
-corresponding `microvm@<name>.service`, matching the behavior of
-ordinary `systemd.services.<name>` — set:
+To opt in to reconciliation, so that removing a VM from `microvm.vms.<name>`
+and running `nixos-rebuild switch` stops the corresponding `microvm@<name>.service`,
+matches the behavior of ordinary `systemd.services.<name>` set:
 
 ```nix
 microvm.host.stopOrphans = true;
@@ -92,10 +89,7 @@ microvm.host.stopOrphans = true;
 
 When enabled:
 
-- State under `/var/lib/microvms/<name>/` is preserved, in case the VM
-  is only being temporarily deactivated. To fully remove a VM including
-  its disk images, `rm -rf /var/lib/microvms/<name>` after the service
-  has stopped.
-- Imperative MicroVMs managed via the `microvm` command are never
-  touched; only VMs that were previously declared in `microvm.vms` are
-  considered.
+- State under `/var/lib/microvms/<name>/` is preserved, in case the VM is only being temporarily deactivated.
+  To fully remove a VM including its disk images, `rm -rf /var/lib/microvms/<name>` after the service has stopped.
+- Imperative MicroVMs managed via the `microvm` command are never touched;
+  only VMs that were previously declared in `microvm.vms` are considered.
--- a/doc/src/microvm-command.md
+++ b/doc/src/microvm-command.md
@@ -64,16 +64,35 @@ ls -l /var/lib/microvms/*/{current,booted}/share/microvm/system
 
 ## Removing MicroVMs
 
+### Imperative MicroVMs
+
 First, stop the MicroVM:
 
 ```bash
 systemctl stop microvm@$NAME
 ```
 
+### Declarative MicroVMs
+
+Remove the VM from `microvm.vms.<name>` in your host configuration and
+stop the VM:
+
+```bash
+systemctl stop microvm@$NAME
+```
+
+Removing a declaration does **not** automatically stop the running
+service by default. To opt in to that behavior, set
+`microvm.host.stopOrphans = true;` on the host — see
+[Declarative MicroVMs → Reconciliation on host rebuild](./declarative.md).
+
+### Cleaning up state
+
 If you don't use absolute filesystem paths for sockets, volumes, or
 shares, all MicroVM state is kept under `/var/lib/microvms/$NAME/`.
-The `microvm@.serivce` systemd service template depends on existence
-of this directory.
+The `microvm@.service` systemd service template depends on existence
+of this directory. State is **not** removed automatically, even for
+declarative MicroVMs — clean it up manually:
 
 ```bash
 rm -rf /var/lib/microvms/$NAME
--- a/doc/src/microvm-command.md
+++ b/doc/src/microvm-command.md
@@ -81,18 +81,17 @@ stop the VM:
 systemctl stop microvm@$NAME
 ```
 
-Removing a declaration does **not** automatically stop the running
-service by default. To opt in to that behavior, set
-`microvm.host.stopOrphans = true;` on the host — see
-[Declarative MicroVMs → Reconciliation on host rebuild](./declarative.md).
+Removing a declaration does **not** automatically stop the running service by default.
+To opt in to that behavior, set `microvm.host.stopOrphans = true;` on the host.
+See [Declarative MicroVMs → Reconciliation on host rebuild](./declarative.md) for more information.
 
 ### Cleaning up state
 
 If you don't use absolute filesystem paths for sockets, volumes, or
 shares, all MicroVM state is kept under `/var/lib/microvms/$NAME/`.
 The `microvm@.service` systemd service template depends on existence
-of this directory. State is **not** removed automatically, even for
-declarative MicroVMs — clean it up manually:
+of this directory. State is **never** removed automatically, even for
+declarative MicroVMs — it must be manually cleaned up:
 
 ```bash
 rm -rf /var/lib/microvms/$NAME
--- a/nixos-modules/host/default.nix
+++ b/nixos-modules/host/default.nix
@@ -312,6 +312,24 @@ in
     # Enable Kernel Same-Page Merging
     hardware.ksm.enable = lib.mkDefault true;
 
+    system.activationScripts.microvm-stop-orphans = lib.optionalString config.microvm.host.stopOrphans ''
+      declared=" ${lib.concatStringsSep " " (builtins.attrNames config.microvm.vms)} "
+      if [ -d /run/current-system/etc/systemd/system ]; then
+        for f in /run/current-system/etc/systemd/system/install-microvm-*.service; do
+          [ -e "$f" ] || continue
+          name="$(basename "$f" .service)"
+          name="''${name#install-microvm-}"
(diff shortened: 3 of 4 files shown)


nixos-modules/host/default.nix
```diff
@@ -312,6 +312,24 @@ in
     # Enable Kernel Same-Page Merging
     hardware.ksm.enable = lib.mkDefault true;
 
+    system.activationScripts.microvm-stop-orphans = lib.optionalString config.microvm.host.stopOrphans ''
+      declared=" ${lib.concatStringsSep " " (builtins.attrNames config.microvm.vms)} "
```
SuperSandro2000:
```suggestion
      declared=" ${lib.concatMapAttrsStringSep " " builtins.attrNames config.microvm.vms} "
```
Could be that args need some changing

nixos-modules/host/default.nix
```diff
@@ -312,6 +312,24 @@ in
...
+      if [ -d /run/current-system/etc/systemd/system ]; then
+        for f in /run/current-system/etc/systemd/system/install-microvm-*.service; do
+          [ -e "$f" ] || continue
+          name="$(basename "$f" .service)"
+          name="''${name#install-microvm-}"
+          case "$declared" in
+            *" $name "*) continue ;;
+          esac
+          if ${pkgs.systemd}/bin/systemctl is-active --quiet "microvm@$name.service" 2>/dev/null; then
```
SuperSandro2000:
```suggestion
          if ${lib.getExe config.systemd.package} is-active --quiet "microvm@$name.service" 2>/dev/null; then
```

nixos-modules/host/default.nix
```diff
@@ -312,6 +312,24 @@ in
...
+            echo "microvm: stopping orphan microvm@$name.service (removed from microvm.vms)" >&2
+            ${pkgs.systemd}/bin/systemctl stop "microvm@$name.service" || true
```
SuperSandro2000:
```suggestion
            ${lib.getExe config.systemd.package} stop "microvm@$name.service" || true
```

doc/src/microvm-command.md
```diff
@@ -64,16 +64,35 @@ ls -l /var/lib/microvms/*/{current,booted}/share/microvm/system
...
+[Declarative MicroVMs → Reconciliation on host rebuild](./declarative.md).
+
+### Cleaning up state
+
 If you don't use absolute filesystem paths for sockets, volumes, or
 shares, all MicroVM state is kept under `/var/lib/microvms/$NAME/`.
-The `microvm@.serivce` systemd service template depends on existence
-of this directory.
+The `microvm@.service` systemd service template depends on existence
+of this directory. State is **not** removed automatically, even for
```
SuperSandro2000:
```suggestion
of this directory. State is **never** removed automatically, even for
```

doc/src/microvm-command.md
```diff
@@ -64,16 +64,35 @@ ls -l /var/lib/microvms/*/{current,booted}/share/microvm/system
...
+declarative MicroVMs — clean it up manually:
```
SuperSandro2000:
```suggestion
declarative MicroVMs — it must be manually cleaned up:
```

doc/src/microvm-command.md
```diff
@@ -64,16 +64,35 @@ ls -l /var/lib/microvms/*/{current,booted}/share/microvm/system
...
+stop the VM:
+
+```bash
+systemctl stop microvm@$NAME
+```
+
+Removing a declaration does **not** automatically stop the running
+service by default. To opt in to that behavior, set
+`microvm.host.stopOrphans = true;` on the host — see
+[Declarative MicroVMs → Reconciliation on host rebuild](./declarative.md).
```
SuperSandro2000:
```suggestion
Removing a declaration does **not** automatically stop the running service by default.
To opt in to that behavior, set `microvm.host.stopOrphans = true;` on the host.
See [Declarative MicroVMs → Reconciliation on host rebuild](./declarative.md) for more information.
```

doc/src/declarative.md
```diff
@@ -71,3 +71,31 @@ microvm.vms = {
...
 closure size of the host's system.
+
+## Reconciliation on host rebuild
+
+By default, removing a VM from `microvm.vms.<name>` does **not** stop
+the running `microvm@<name>.service`. Out of the box, MicroVM
+lifecycle is decoupled from host configuration — updates are applied
+via the imperative [`microvm` command](./microvm-command.md), and
+removing a declaration simply stops managing the VM declaratively
+without disturbing the running process.
```
SuperSandro2000:
```suggestion
By default, removing a VM from `microvm.vms.<name>` does **not** stop the running `microvm@<name>.service`,
because MicroVM lifecycle is decoupled from host configuration.
Updates are meant to be applied via the imperative [`microvm` command](./microvm-command.md),
and removing a declaration simply stops managing the VM declaratively.
```
Lets do one sentence per line for new docs. 80 line lengths are not easy to read for dyslexic people and also make diffs harder.

doc/src/declarative.md
```diff
@@ -71,3 +71,31 @@ microvm.vms = {
...
+
+To opt in to reconciliation — so that removing a VM from
+`microvm.vms.<name>` and running `nixos-rebuild switch` stops the
+corresponding `microvm@<name>.service`, matching the behavior of
+ordinary `systemd.services.<name>` — set:
```
SuperSandro2000:
```suggestion
To opt in to reconciliation, so that removing a VM from `microvm.vms.<name>`
and running `nixos-rebuild switch` stops the corresponding `microvm@<name>.service`,
matches the behavior of ordinary `systemd.services.<name>` set:
```

doc/src/declarative.md
```diff
@@ -71,3 +71,31 @@ microvm.vms = {
...
+
+When enabled:
+
+- State under `/var/lib/microvms/<name>/` is preserved, in case the VM
+  is only being temporarily deactivated. To fully remove a VM including
+  its disk images, `rm -rf /var/lib/microvms/<name>` after the service
+  has stopped.
+- Imperative MicroVMs managed via the `microvm` command are never
+  touched; only VMs that were previously declared in `microvm.vms` are
+  considered.
```
SuperSandro2000:
```suggestion
- Imperative MicroVMs managed via the `microvm` command are never touched;
  only VMs that were previously declared in `microvm.vms` are considered.
```

doc/src/declarative.md
```diff
@@ -71,3 +71,31 @@ microvm.vms = {
...
+```nix
+microvm.host.stopOrphans = true;
+```
+
+When enabled:
+
+- State under `/var/lib/microvms/<name>/` is preserved, in case the VM
+  is only being temporarily deactivated. To fully remove a VM including
+  its disk images, `rm -rf /var/lib/microvms/<name>` after the service
+  has stopped.
```
SuperSandro2000:
```suggestion

- State under `/var/lib/microvms/<name>/` is preserved, in case the VM is only being temporarily deactivated.
  To fully remove a VM including its disk images, `rm -rf /var/lib/microvms/<name>` after the service has stopped.
```

brandonros:
Thanks for the review @SuperSandro2000! All ten suggestions applied in 2f61b71 with one minor correction:

- **systemd → systemctl lookup**: used `lib.getExe' config.systemd.package "systemctl"` rather than `lib.getExe config.systemd.package`. `getExe` falls back to `${systemd}/bin/systemd` (not `systemctl`) because the systemd package has no `meta.mainProgram` set — locally I confirmed this with a deprecation warning. `getExe'` lets us specify the binary explicitly, which matches what's actually wanted here.
- `concatMapAttrsStringSep`: used `(name: _: name)` as the mapper function so it correctly picks out the attribute names.
- All doc suggestions applied verbatim; the rest of the new sections reformatted to one sentence per line per the style rule.

nixos-modules/host/default.nix
```diff
@@ -312,6 +312,24 @@ in
     # Enable Kernel Same-Page Merging
     hardware.ksm.enable = lib.mkDefault true;
 
+    system.activationScripts.microvm-stop-orphans = lib.optionalString config.microvm.host.stopOrphans ''
+      declared=" ${lib.concatMapAttrsStringSep " " (name: _: name) config.microvm.vms} "
+      if [ -d /run/current-system/etc/systemd/system ]; then
```
astro: Why use `/run/current-system/etc`, not `/etc`?

nixos-modules/host/default.nix
```diff
@@ -312,6 +312,24 @@ in
...
+        for f in /run/current-system/etc/systemd/system/install-microvm-*.service; do
+          [ -e "$f" ] || continue
+          name="$(basename "$f" .service)"
+          name="''${name#install-microvm-}"
+          case "$declared" in
+            *" $name "*) continue ;;
+          esac
+          if ${lib.getExe' config.systemd.package "systemctl"} is-active --quiet "microvm@$name.service" 2>/dev/null; then
```
astro: I think we can assume `systemctl` to be in `$PATH`.

nixos-modules/host/default.nix
brandonros: Good catch — removed. By the time our activation script runs, `/etc` already reflects the new generation (etc activation precedes ours), so `/etc/systemd/system/install-microvm-*.service` would only show the *currently-declared* VMs and couldn't reveal orphans. The original code reached into `/run/current-system/etc` specifically to see the *previous* generation. In a2de3b2 I switched to `systemctl list-unit-files 'install-microvm-*.service'` instead — systemd hasn't been told to `daemon-reload` yet during activation, so its in-memory state still reflects the old gen. No filesystem introspection needed.

nixos-modules/host/default.nix
brandonros: Agreed — switched to bare `systemctl` in a2de3b2.

brandonros:
Thanks @astro — both addressed in a2de3b2. The activation script now uses `systemctl list-unit-files 'install-microvm-*.service'` to find previously-declarative VMs (systemd's in-memory state is still pre-`daemon-reload` during activation scripts, so it reflects the old generation without needing to touch `/run/current-system`), and uses bare `systemctl` throughout. Net diff vs prior commit: -14/+11 lines — the new version is meaningfully simpler. Ready for another look whenever you have a moment.

nixos-modules/host/default.nix
```diff
@@ -312,6 +312,21 @@ in
     # Enable Kernel Same-Page Merging
     hardware.ksm.enable = lib.mkDefault true;
 
+    system.activationScripts.microvm-stop-orphans = lib.optionalString config.microvm.host.stopOrphans ''
+      declared=" ${lib.concatMapAttrsStringSep " " (name: _: name) config.microvm.vms} "
+      for unit in $(systemctl list-unit-files --no-legend 'install-microvm-*.service' 2>/dev/null | awk '{print $1}'); do
```
karaolidis: The problem with LLM-generated code and PRs, such as this one, is that they don't work:

```
...
/nix/store/a0kzb6wmdkp6h9f6z6cs367iw7f44f6f-nixos-system-jupiter-26.05.19700101.dirty/activate: line 130: awk: command not found
...
```

brandonros: let's see if https://github.com/microvm-nix/microvm.nix/pull/509/commits/a6c36902e2345772c903a6475e337949b72b3272 fixes it

PR #510, opened 2026-04-22
vfkit: use `virtualisation.rosetta` instead of manual set up

r4v3n6101:
Replace manual binfmt configuration with the nixpkgs-provided option. This gives more flexible setup compared to the previous implementation.

### DIFF
--- a/nixos-modules/microvm/rosetta.nix
+++ b/nixos-modules/microvm/rosetta.nix
@@ -1,19 +1,4 @@
-{ config, lib, pkgs, ... }:
-
-let
-  cfg = config.microvm.vfkit.rosetta;
-  mountPoint = "/run/rosetta";
-  mountTag = "rosetta";
-in
-lib.mkIf (config.microvm.hypervisor == "vfkit" && cfg.enable) {
-  fileSystems.${mountPoint} = {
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
Please inline the let cfg in that it is now only used once.

nixos-modules/microvm/rosetta.nix
```diff
@@ -1,19 +1,4 @@
...
-    device = mountTag;
-    fsType = "virtiofs";
-  };
-
-  boot.binfmt.registrations.rosetta = {
-    interpreter = "${mountPoint}/rosetta";
-    magicOrExtension = ''\x7fELF\x02\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x02\x00\x3e\x00'';
-    mask = ''\xff\xff\xff\xff\xff\xfe\xfe\x00\xff\xff\xff\xff\xff\xff\xff\xff\xfe\xff\xff\xff'';
-  };
+{ config, lib, ... }:
```
SuperSandro2000, line 1:
```suggestion
{ config, lib, ... }:

```

SuperSandro2000 (approved):

PR #519, opened 2026-05-09
kvmtool runner puts the literal "image" as the source for volumes when provided as configuration

kvnvelasco:
Looks like this was probably a typo. Found when testing KVM machines


astro:
Thanks!

PR #527, opened 2026-05-27
fix(qemu): fix memory hotplug config

### LINKED ISSUE #514: QEMU virtio-mem support is broken
The changes in #490 aren't sufficient for virtio-mem support with QEMU.

To be able to make use of memory hotplugging, the `-m` flag needs at least the `maxmem` option, which should be `mem + hotplugMem` (aligned to the page size). The documentation says

> If slots and maxmem are not specified, memory hotplug won’t be enabled and the guest startup RAM will never increase.

but for me it worked with just maxmem.

Furthermore, QEMU's microvm machine model does not support virtio-mem, or hotplugging of any kind, as per the [the documentation](https://www.qemu.org/docs/master/system/i386/microvm.html). Trying to use virtio-mem anyway will try to map the device at address 0x0, which fails:

```
[    7.795674] virtio_mem virtio2: start address: 0x0
[    7.795678] virtio_mem virtio2: region size: 0x80000000
[    7.795679] virtio_mem virtio2: device block size: 0x200000
[    7.795680] virtio_mem virtio2: nid: 0
[    7.795681] virtio_mem virtio2: memory block size: 0x8000000
[    7.795682] virtio_mem virtio2: subblock size: 0x200000
[    7.795684] virtio_mem virtio2: could not reserve device region
[    7.795685] virtio_mem virtio2: reloading the driver is not supported
[    7.810210] virtio_mem virtio2: probe with driver virtio_mem failed with error -16
```

If one tries to manually set the `memaddr` of the `virtio-mem-pci` device, one of several errors occur. If one sets `maxmem` correctly, trying to set the address to 1 Gi results in a `kernel NULL pointer dereference` and setting it to 4 Gi results in a `KVM internal error. Suberror: 3` (on my system; the list of errors is probably not exhaustive). Without the `maxmem` fix, I've been getting `Oops: general protection fault, probably for non-canonical address 0xe172214a17183636` instead.

It's worth noting that with the correct machine type, QEMU actively complains about the lack of `maxmem`. With the `microvm` machine type, it just happily breaks in the aforementioned ways.

kajusnau:
Added missing maxmem option when hotplugMem != 0
Added explicit error when machine type is microvm and hotplugMem != 0

QEMU requires specifying the maxmem option to use virtio-mem
The microvm machine type does not support virtio-mem

Closes #514

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -162,13 +162,15 @@ lib.warnIf (mem == 2048) ''
   inherit tapMultiQueue;
 
   command = if initialBalloonMem != 0
-  then throw "qemu does not support initialBalloonMem"
+  then throw "QEMU does not support initialBalloonMem"
+  else if (useHotPlugMemory && machine == "microvm")
+  then throw "QEMU's microvm machine type does not support virtio-mem"
   else lib.escapeShellArgs (
     [
       "${qemu}/bin/qemu-system-${arch}"
       "-name" hostName
       "-M" machineConfig
-      "-m" (toString mem)
+      "-m" "${toString mem}M${lib.optionalString useHotPlugMemory ",maxmem=${toString (mem + hotplugMem)}M"}"
       "-smp" (toString vcpu)
       "-nodefaults" "-no-user-config"
       # qemu just hangs after shutdown, allow to exit by rebooting


dominikh:
I've tested it and it looks to be working.

PR #537, opened 2026-06-18
Fix virtiofsd panic when hotplugging memory

0xADD1E:
As noted in https://gitlab.com/virtio-fs/virtiofsd/-/work_items/165 when using hotplug memory, if not configured correctly will result in

```
virtiofsd-ro-store called `Result::unwrap()` on an `Err` value: QueueWriter(FindMemoryRegion)
```

Ensures memory is shared


0xADD1E:
There's almost certainly a better way of scoping this but my coffee hasn't kicked in enough for me to think through it yet. It does seem to resolve the issue though (at least, when using virtiofs and hotplug mem; might break other use cases)

astro (approved):

karaolidis (approved):
Also +1ing this, I was experiencing the mentioned issue, and this resolves it :)
Thanks!

astro:
@0xADD1E Could you please get this ready for review?

PR #555, opened 2026-07-13
Fix the executable path in run-waypipe

mksafavi:
The bin path was added twice:
```shell
[vm@nixos:~]$ run-waypipe 
/run/current-system/sw/bin/run-waypipe: line 2: /nix/store/8gzmq3wh3sig66mbns019qz96ab0cyh3-waypipe-0.11.0/bin/waypipe/bin/waypipe: Not a directory
```


astro:
Thank you!

PR #562, opened 2026-07-16
Drop x86_64-darwin from systems

mksafavi:
CI fails to build on main because `x86_64-darwin` support has ended.
Opened this as draft as there are other places that `x86_64-darwin` is used that could be removed.
I also wasn't sure if you want to drop `x86_64-darwin` or pin the input.


SuperSandro2000:
Here we usually take the escape forward. It's not maintainable to keep supporting x86_64-darwin and it also frees a little bit of compute.

SuperSandro2000 (approved):

PR #563, opened 2026-07-18
virtiofs: add per-share posixAcl and extraArgs

luochen1990:
virtiofsd supports UID/GID remapping via `--translate-uid`/`--translate-gid`, but those conflict with `--posix-acl`, which microvm.nix hard-codes for every virtiofs share. The global `microvm.virtiofsd.extraArgs` is also insufficient because it applies to all shares. This PR makes both flags per-share, so an individual share can opt out of `--posix-acl` and pass `--translate-uid`.

Two new attributes on each `virtiofs` share entry (naming follows the existing `readOnly` style — direct noun, not `enableX`):

- `posixAcl` (bool, default `true`): toggle `--posix-acl --xattr`
- `extraArgs` (list of strings, default `[]`): per-share virtiofsd args, appended after the global `microvm.virtiofsd.extraArgs`

Both default to values that produce identical virtiofsd argv for existing configs. See `doc/src/shares.md` "Per-share virtiofsd options" for a usage example (UID/GID remapping across host/guest boundary).

Verified by inspecting the generated supervisord config for two shares in the same VM: default share still emits `--posix-acl --xattr` (unchanged), and a share with `posixAcl = false; extraArgs = ["--translate-uid" "guest:999:1000:1"]` correctly emits `--translate-uid` without `--posix-acl`. Also confirmed `virtiofsd` itself starts successfully with that invocation. Added a `virtiofs` variant to `checks/default.nix` exercising both new options.

### DIFF
--- a/nixos-modules/microvm/options.nix
+++ b/nixos-modules/microvm/options.nix
@@ -420,7 +420,8 @@ in
             default = true;
             description = ''
               Pass `--posix-acl --xattr` to virtiofsd. Disable when using
-              `--translate-uid`/`--translate-gid` (they conflict with `--posix-acl`).
+              `--translate-uid`/`--translate-gid` (asserted to be mutually exclusive
+              with `--posix-acl`, see asserts.nix).
             '';
           };
           extraArgs = mkOption {
--- a/nixos-modules/microvm/options.nix
+++ b/nixos-modules/microvm/options.nix
@@ -415,6 +415,19 @@ in
             description = "Virtiofs caching policy for the file system, ignored when 9p is used";
             default = "auto";
           };
+          posixAcl = mkOption {
+            type = bool;
+            default = true;
+            description = ''
+              Pass `--posix-acl --xattr` to virtiofsd. Disable when using
+              `--translate-uid`/`--translate-gid` (they conflict with `--posix-acl`).
+            '';
+          };
+          extraArgs = mkOption {
+            type = listOf str;
+            default = [];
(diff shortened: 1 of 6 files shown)


nixos-modules/microvm/options.nix
```diff
@@ -415,6 +415,19 @@ in
             description = "Virtiofs caching policy for the file system, ignored when 9p is used";
             default = "auto";
           };
+          posixAcl = mkOption {
+            type = bool;
+            default = true;
+            description = ''
+              Pass `--posix-acl --xattr` to virtiofsd. Disable when using
+              `--translate-uid`/`--translate-gid` (they conflict with `--posix-acl`).
```
SuperSandro2000: Can we assert that?

luochen1990: Yes, let me work on this -- adding assertions.

luochen1990: Done — added the assertion in commit b680c7c. Since each virtiofs share runs its own virtiofsd, the conflict is per-share, but `--translate-uid`/`--translate-gid` can come from either the per-share `extraArgs` (added in this PR) or the global `microvm.virtiofsd.extraArgs`. The assertion checks both sources against each share's `posixAcl`, and points users to set `posixAcl = false` in the error message.

Separate note: the failing `cloud-hypervisor-graphics` check is pre-existing on `main` — spectrum's vhost patches need updating for cloud-hypervisor 52.0. Unrelated to this PR.

SuperSandro2000 (approved):

PR #565, opened 2026-07-21
Add QEMU virtio-gpu Vulkan options

mksafavi:
Adds support for vulkan translation layers on QEMU.
https://www.qemu.org/docs/master/system/devices/virtio/virtio-gpu.html

### DIFF
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -239,15 +239,20 @@ lib.warnIf (mem == 2048) ''
     ] ++
        (if graphics.enable then (
        let
+         vulkanArgs = lib.optionals (graphics.vulkan != null) [
+          "blob=true"
+          "hostmem=${graphics.hostmem}"
+          "${graphics.vulkan}=on"
+         ];
          displayArgs = {
            cocoa = [
              "-display" "cocoa" "-device" "virtio-gpu"
            ];
            gtk = [
-             "-display" "gtk,gl=on" "-device" "virtio-vga-gl"
+             "-display" "gtk,gl=on" "-device" (lib.concatStringsSep "," ([ "virtio-vga-gl" ] ++ vulkanArgs))
            ];
            headless = [
-             "-display" "egl-headless" "-device" "virtio-gpu-gl"
+             "-display" "egl-headless" "-device" (lib.concatStringsSep "," ([ "virtio-gpu-gl" ] ++ vulkanArgs))
            ];
          }.${graphics.backend};
        in
--- a/nixos-modules/microvm/options.nix
+++ b/nixos-modules/microvm/options.nix
@@ -617,6 +617,23 @@ in
           Path of vhost-user socket
         '';
       };
+
+      vulkan = mkOption {
+        type = with types; nullOr (enum [ "venus" "drm_native_context" ]);
+        default = null;
+        description = ''
+          QEMU Vulkan translation protocol to use for enabling virtio-gpu Vulkan support.
+        '';
+      };
+
+      hostmem = mkOption {
+        type = types.str;
+        default = "256M";
+        description = ''
+          Size of the QEMU virtio-gpu host memory window, typically between 256M and 8G.
+          The value must include a unit suffix (M or G).
+        '';
+      };
     };
 
     vmHostPackages = mkOption {


mksafavi:
I'm not sure about the types here. For example, `hostmem` can be set to `null` while `vulkan` is enabled. Should I add an assertion, or remove the `null` type from `hostmem` altogether? I'm sure it could be modeled better, but this is the smallest diff I could make.

`blob=true` could also become an option, it enables OpenGL 4.6 features. But I couldn't find a use case for enabling it on its own. Let me know what you think

lib/runners/qemu.nix
```diff
@@ -239,15 +239,24 @@ lib.warnIf (mem == 2048) ''
...
+               drm_native_context = ",blob=true,hostmem=${graphics.hostmem},drm_native_context=on";
+             }
+             .${graphics.vulkan};
          displayArgs = {
            cocoa = [
              "-display" "cocoa" "-device" "virtio-gpu"
            ];
            gtk = [
-             "-display" "gtk,gl=on" "-device" "virtio-vga-gl"
+             "-display" "gtk,gl=on" "-device" "virtio-vga-gl${vulkanArgs}"
```
SuperSandro2000: Can we make the last argument an array and join that with `,`?

SuperSandro2000:
> `blob=true` could also become an option, it enables OpenGL 4.6 features. But I couldn't find a use case for enabling it on its own. Let me know what you think

Then we do not add an option for it for now. We don't need an option for literally everything.

lib/runners/qemu.nix
```diff
@@ -239,15 +239,24 @@ lib.warnIf (mem == 2048) ''
...
        let
+         vulkanArgs =
+           if graphics.vulkan == null then
+             ""
+           else
+             {
+               venus = ",blob=true,hostmem=${graphics.hostmem},venus=true";
+               drm_native_context = ",blob=true,hostmem=${graphics.hostmem},drm_native_context=on";
+             }
+             .${graphics.vulkan};
```
```suggestion
           if graphics.vulkan == null then
             ""
           else
             {
               venus = [ "blob=true" "hostmem=${graphics.hostmem}" "venus=true" ];
               drm_native_context = [ "blob=true" "hostmem=${graphics.hostmem}" "drm_native_context=on" ];
             }
             .${graphics.vulkan};
```

nixos-modules/microvm/options.nix
```diff
@@ -617,6 +617,24 @@ in
...
+      vulkan = mkOption {
+        type = with types; nullOr (enum [ "venus" "drm_native_context" ]);
+        default = null;
+        description = ''
+          QEMU Vulkan translation protocol to use for enabling virtio-gpu Vulkan support.
+        '';
+      };
+
+      hostmem = mkOption {
+        type = with types; nullOr str;
```
SuperSandro2000: For what do we need null here? Could we just use vulkan != null to avoid the null type here?

mksafavi: I guess I was overthinking it. When `vulkan == null`, `hostmem` is just a value that's not used anywhere. There's no need to keep the null 👍

lib/runners/qemu.nix
```diff
@@ -238,15 +239,24 @@ lib.warnIf (mem == 2048) ''
...
        (if graphics.enable then (
        let
+         vulkanArgs = lib.optionals (graphics.vulkan != null ) [
+          "blob=true"
+          "hostmem=${graphics.hostmem}"
+          {
+            venus = "venus=true";
+            drm_native_context = "drm_native_context=on";
+          }
+          .${graphics.vulkan}
```
mksafavi: Does QEMU accept `true/false` and `on/off` interchangeably? 
If so, we can replace that section with this:
```suggestion
          "${graphics.vulkan}=on"
```

mksafavi: It guess the virtio-gpu documentation was outdated. qemu cli interface only lists on/off for booleans:
```shell
❯ qemu-system-x86_64 -device virtio-gpu-gl,help | grep -E "blob|venus|drm_native_context"
  blob=<bool>            - on/off (default: off)
  drm_native_context=<bool> - on/off (default: off)
  venus=<bool>           - on/off (default: off)
```

SuperSandro2000 (approved):

PR #572, opened 2026-07-31
Fix SSH over vsock when no virtio-PCI devices are available

### LINKED ISSUE #512: Explicitly set `systemd.ssh_listen` for ssh vsock support
According to https://github.com/yuwata/systemd/blob/f0483f308a4daed188deee776aa7a7a733293642/src/core/kmod-setup.c#L83, systemd sets up the vsock listener only when it sees a vsock PCI device. This means if devType is not `pci`, https://github.com/microvm-nix/microvm.nix/blob/2f2f62fdfdca2750e3399f66bd03986ab967e5ca/lib/runners/qemu.nix#L345 sets it up as a normal device, and systemd won't detect it automatically.

A more reliable way would be to configure ssh generator explicitly via kernel parameters `systemd.ssh_listen=vsock::22` (https://www.freedesktop.org/software/systemd/man/latest/systemd-ssh-generator.html)

mksafavi:
SSH over `vsock` stops working when I disable the only `virtiofs` share in my example VM (the read-only nix store).
I think `systemd-ssh-generator` fails when a `virtio-pci` device is not present. I'm not sure if this is an upstream issue, but what weird bug :)
Either way, adding `vmw_vsock_virtio_transport` to initrd available kernel modules fixes the issue.

Closes https://github.com/microvm-nix/microvm.nix/issues/512

Before:
```
[vm@nixos:~]$ ss -lnp | grep :22
tcp   LISTEN 0      128                                                   0.0.0.0:22               0.0.0.0:*    
tcp   LISTEN 0      128                                                      [::]:22                  [::]:*    
```
After:
```
[vm@nixos:~]$ ss -lnp | grep :22
tcp   LISTEN 0      128                                                   0.0.0.0:22               0.0.0.0:*         
tcp   LISTEN 0      128                                                      [::]:22                  [::]:*         
v_str LISTEN 0      0                                                           *:22                     *:*         
```

### DIFF
--- a/nixos-modules/microvm/system.nix
+++ b/nixos-modules/microvm/system.nix
@@ -10,7 +10,6 @@
 
 
     boot.loader.grub.enable = false;
-    # boot.initrd.systemd.enable = lib.mkDefault true;
     boot.initrd.kernelModules = [
       "virtio_mmio"
       "virtio_pci"
--- a/nixos-modules/microvm/system.nix
+++ b/nixos-modules/microvm/system.nix
@@ -27,6 +27,11 @@
       "overlay"
     ];
 
+    boot.initrd.availableKernelModules = [
+      # Fix SSH over vsock when no virtio-PCI devices are available
+      "vmw_vsock_virtio_transport"
+    ];
+
     microvm.kernelParams = let
       # When a store disk is used, we can drop references to the packed contents as the squashfs/erofs contains all paths.
       toplevel = if config.microvm.storeOnDisk then


nixos-modules/microvm/system.nix
```diff
@@ -10,7 +10,6 @@
 
 
     boot.loader.grub.enable = false;
-    # boot.initrd.systemd.enable = lib.mkDefault true;
```
mksafavi, line 13: Do you want me to drop or squash this?

SuperSandro2000: Fine like it is :)

nixos-modules/microvm/system.nix
```diff
@@ -28,6 +27,10 @@
       "overlay"
     ];
 
+    boot.initrd.availableKernelModules = [
+      "vmw_vsock_virtio_transport"
```
SuperSandro2000, line 32:
```suggestion
      # Fix SSH over vsock when no virtio-PCI devices are available
      "vmw_vsock_virtio_transport"
```

PR #584, opened 2026-08-15
fix(crosvm): attach store disk as non-root

vadika:
## What changed

Attach the generated store image to Crosvm with `--block <image>,ro=true` instead of `-r`.

## Why

Crosvm interprets `-r` as the guest root disk and appends `root=/dev/vda`. For `microvm.storeOnDisk`, the generated store image is a read-only Nix store input and must not replace the guest root filesystem.

Signed-off-by: vadik likholetov <vadikas@gmail.com>

### DIFF
--- a/lib/runners/crosvm.nix
+++ b/lib/runners/crosvm.nix
@@ -74,7 +74,7 @@ in {
       )
       ++
       lib.optionals storeOnDisk [
-        "-r" storeDisk
+        "--block" "${storeDisk},ro=true"
       ]
       ++
       lib.optionals graphics.enable [


vadika:
Rebased the single fix commit onto current `main` (`71beea0`); new head is `b48464e`.

Focused validation passed with a direct build of `checks/crosvm-store-disk.nix`: the generated runner uses `--block <storeDisk>,ro=true` and emits no Crosvm root-disk flag. The remaining Gradient status is an approval gate.

checks/crosvm-store-disk.nix
```diff
@@ -0,0 +1,36 @@
...
+    nativeBuildInputs = [ pkgs.gnugrep ];
+  } ''
+    runner=${microvm.config.microvm.declaredRunner}/bin/microvm-run
+    store_disk=${microvm.config.microvm.storeDisk}
+
+    grep -F -- "--block '$store_disk,ro=true'" "$runner"
+    if grep -Eq -- '(^| )(-r|--root) ' "$runner"; then
+      echo "Crosvm runner incorrectly marks the store disk as the guest root" >&2
+      exit 1
+    fi
```
SuperSandro2000: We are just greping for a line here. I am not fully into the tests but if we are not booting the VM and checking the mount point in it, than we can also ommit the test.

astro (approved):

PR #586, opened 2026-08-24
crosvm: Support platform passthrough layouts

vadika:
## Summary

- accept platform devices in microvm.devices for Crosvm
- expose Crosvm VFIO IOMMU, guest-address, device-tree symbol, fixed-MMIO, and early-mapping options
- pass device-tree overlays and platform VFIO devices to the runner
- expose typed AArch64 RAM-base and platform-MMIO aperture settings
- preserve the existing runner command when no explicit layout is configured
- reject invalid hypervisor, architecture, bus, missing-symbol, missing-overlay, and overlapping-layout combinations
- add focused runner and configuration coverage

## Verification

- git diff --check
- nix build --accept-flake-config .#hydraJobs.x86_64-linux.crosvm-platform-devices --no-link --print-build-logs

### DIFF
--- a/lib/runners/crosvm.nix
+++ b/lib/runners/crosvm.nix
@@ -13,9 +13,24 @@ let
     socket devices vsock graphics credentialFiles
     kernel initrdPath storeDisk storeOnDisk;
   inherit (microvmConfig.crosvm)
-    pivotRoot extraArgs deviceTreeOverlays memoryBase platformMmio;
+    pivotRoot extraArgs deviceTreeOverlays memoryBase platformMmio protection;
 
   crosvmPkg = microvmConfig.crosvm.package;
+  isProtected = protection.mode != "unprotected";
+  protectionArgs =
+    {
+      unprotected = [ ];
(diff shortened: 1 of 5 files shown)


lib/runners/crosvm.nix
```diff
@@ -27,6 +27,15 @@ let
     vulkan = true;
   };
 
+  vfioDeviceArgs = { bus, path, crosvm, ... }: [
+    "--vfio"
+    ({
+      pci = "/sys/bus/pci/devices/${path},iommu=${crosvm.iommu}${lib.optionalString (crosvm.guestAddress != null) ",guest-address=${crosvm.guestAddress}"}${lib.optionalString (crosvm.dtSymbol != null) ",dt-symbol=${crosvm.dtSymbol}"}";
```
SuperSandro2000, line 51: Please use a list that is joind with , so that we do not need to care about commas in optionalStrings.

SuperSandro2000 (commented):
That's a lot of stuff and I must admit I do not understand most of it... I am not sure if this is a good fit for the project @astro what do you say?

PR #592, opened 2026-09-01
feat: open the store disk with O_DIRECT where the hypervisor supports it

maxdjohnson:
## Summary

`microvm.volumes.*.direct` already lets a volume bypass the host page cache, but the generated `microvm.storeDisk` was always opened buffered. This adds `microvm.storeDiskDirect` (default `true` on supported VMs, otherwise `false`) and passes it through for every hypervisor that takes a per-disk direct-I/O flag:

| Hypervisor | Store disk argument when enabled |
|---|---|
| crosvm | `-r <disk>,o_direct=true` |
| qemu | `-drive id=store,...,cache=none` |
| cloud-hypervisor | `--disk path=<disk>,readonly=on,direct=on` |
| kvmtool | `-d <disk>,direct,ro` |
| stratovirt | `-drive id=store,...,direct=true` (was hard-coded `direct=false`) |
| firecracker, alioth, vfkit | error |

**Behavior change:** existing configurations on the five supported hypervisors now open the store disk with O_DIRECT. Set `microvm.storeDiskDirect = false` to restore buffered reads, e.g. if `/nix/store` is on a filesystem without O_DIRECT support. Noted in CHANGELOG.md.

## Motivation

The store disk is the one image a MicroVM cannot open with direct I/O. Without direct I/O, each block the guest reads from it is cached twice - once in the host page cache and once in the guest's own page cache, which is wasteful esp. on small hosts.

## Testing

- Evaluated the runner command for each hypervisor with the option on and off from a minimal `nixosSystem`; the only differences are the arguments in the table above, and firecracker, alioth, and vfkit raise an error.
- Tested end-to-end on my nixos box running a crosvm guest

### DIFF
--- a/lib/runners/alioth.nix
+++ b/lib/runners/alioth.nix
@@ -13,7 +13,7 @@ let
     user
     vcpu mem balloon initialBalloonMem hotplugMem hotpluggedMem interfaces volumes shares devices vsock
     kernel initrdPath
-    storeDisk storeOnDisk credentialFiles;
+    storeDisk storeOnDisk storeDiskDirect credentialFiles;
 in {
   command =
     if user != null
@@ -28,6 +28,8 @@ in {
     then throw "alioth does not support hotpluggedMem"
     else if credentialFiles != {}
     then throw "alioth does not support credentialFiles"
+    else if storeDiskDirect
+    then throw "alioth does not support storeDiskDirect"
     else builtins.concatStringsSep " " (
       [
         "${aliothPkg}/bin/alioth" "run"
--- a/lib/runners/firecracker.nix
+++ b/lib/runners/firecracker.nix
@@ -12,7 +12,7 @@ let
     vcpu mem balloon initialBalloonMem hotplugMem hotpluggedMem
     interfaces volumes shares devices
     kernel initrdPath
-    storeDisk credentialFiles vsock;
+    storeDisk storeDiskDirect credentialFiles vsock;
   inherit (microvmConfig.firecracker) cpu;
 
   kernelPath = {
@@ -98,6 +98,8 @@ in {
     then throw "hotpluggedMem not implemented for Firecracker"
     else if credentialFiles != {}
     then throw "credentialFiles are not implemented for Firecracker"
+    else if storeDiskDirect
+    then throw "storeDiskDirect is not implemented for Firecracker"
     else lib.escapeShellArgs ([
       "${firecrackerPkg}/bin/firecracker"
       "--config-file" configFile
--- a/lib/runners/vfkit.nix
+++ b/lib/runners/vfkit.nix
@@ -14,7 +14,7 @@ let
 
   inherit (microvmConfig)
     vcpu mem user interfaces shares socket
-    storeOnDisk storeDisk kernel initrdPath kernelParams
+    storeOnDisk storeDisk storeDiskDirect kernel initrdPath kernelParams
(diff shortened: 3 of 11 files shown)


nixos-modules/microvm/options.nix
```diff
@@ -1058,6 +1058,17 @@ in
       default = [ "-c" "zstd" "-j" "$NIX_BUILD_CORES" ];
     };
 
+    storeDiskDirect = mkOption {
+      type = types.bool;
+      default = false;
```
astro: Default to `true`?

astro (approved):

nixos-modules/microvm/options.nix
maxdjohnson: this is a behavior change; but I agree it's a better default. updated to default true on supported VMs, and changelog to note the behavior change

nixos-modules/microvm/options.nix
```diff
@@ -1058,6 +1058,22 @@ in
       default = [ "-c" "zstd" "-j" "$NIX_BUILD_CORES" ];
     };
 
+    storeDiskDirect = mkOption {
+      type = types.bool;
+      default = builtins.elem config.microvm.hypervisor self-lib.hypervisorsWithStoreDiskDirect;
+      defaultText = lib.literalExpression ''builtins.elem config.microvm.hypervisor [ "crosvm" "qemu" "cloud-hypervisor" "kvmtool" "stratovirt" ]'';
```
SuperSandro2000: Nobody is going to update that list in two places. We only want it in one place.

nixos-modules/microvm/options.nix
```diff
@@ -1058,6 +1058,22 @@ in
...
+      description = ''
+        Whether to open the `microvm.storeDisk` image with O_DIRECT,
+        bypassing the host page cache like `microvm.volumes.*.direct`.
+        The guest caches the store disk itself, so the host copy is
+        normally redundant.
+
+        Enabled by default on the hypervisors that support it: crosvm,
```
SuperSandro2000: Another repetition here

lib/runners/vfkit.nix
```diff
@@ -44,9 +44,11 @@ let
     ] else [
       "--device" "virtio-serial,stdio"
     ])
-    ++ lib.optionals storeOnDisk [
+    ++ lib.optionals storeOnDisk (lib.warnIf storeDiskDirect ''
```
SuperSandro2000: This should be a throw like the others

lib/runners/firecracker.nix
```diff
@@ -34,13 +34,15 @@ let
       # Enabling simultaneous multithreading is not supported on aarch64
       smt = system != "aarch64-linux";
     };
-    drives = [ {
+    drives = [ (lib.warnIf storeDiskDirect ''
```
SuperSandro2000: This should also be a throw

lib/runners/alioth.nix
```diff
@@ -39,9 +39,11 @@ in {
         "--entropy"
       ]
       ++
-      lib.optionals storeOnDisk [
+      lib.optionals storeOnDisk (lib.warnIf storeDiskDirect ''
```
SuperSandro2000: Also a throw

nixos-modules/microvm/options.nix
maxdjohnson: good call - updated

PR #598, opened 2026-09-09
feat: add DAX support for alioth and crosvm 

kajusnau:
Added DAX support for virtiofs shares

Opt-in `dax` per virtiofs share, on the two hypervisors that actually implement it.
The guest mounts with `-o dax=always`, so it fails loudly rather than silently degrading if there is no DAX window.

1. **Added DAX support for crosvm**
   virtiofsd can't negotiate a DAX window, so the share goes to crosvm's built-in virtio-fs device instead: no socket, no virtiofsd, no `readOnly`.
   crosvm's window is a fixed 8GiB, so it ignores `daxWindowSize`.

2. **Added DAX support for alioth**
   DAX works on the existing device virtiofsd device and honors `daxWindowSize`.

3. **Added a `virtiofs-dax` check variant**
   Boots a VM with a DAX `/nix/store` share on both hypervisors.
   `dax=always` makes the kernel refuse the mount without a real window, so passing proves it works.

4. **Documentation**
   Added docs covering the new options, the crosvm caveats, and why the other hypervisors are absent.
   Added DAX note to changelog

## Hypervisor support

| Hypervisor | DAX | Notes |
|---|---|---|
| alioth | yes | configurable window |
| crosvm | yes | fixed 8GiB window |
| cloud-hypervisor | no | `docs/fs.md`: "not available"; `--fs` rejects `dax` |
| qemu | no | dropped `cache-size` on `vhost-user-fs-pci` |
| stratovirt, vfkit | no | no DAX at all |
| firecracker, kvmtool | n/a | no virtiofs shares |

Some refs:

- https://virtio-fs.gitlab.io/design.html
- https://lwn.net/Articles/813807/
- https://www.kernel.org/doc/html/latest/filesystems/dax.html

### DIFF
--- a/lib/runners/alioth.nix
+++ b/lib/runners/alioth.nix
@@ -58,10 +58,17 @@ in {
           ]
       ) volumes
       ++
-      builtins.concatMap ({ proto, socket, tag, ... }:
+      builtins.concatMap ({ proto, socket, tag, dax, daxWindowSize, ... }:
         if proto == "virtiofs"
         then [
-          "--fs" (lib.escapeShellArg "vu,socket=${socket},tag=${tag}")
+          "--fs" (lib.escapeShellArg (lib.concatStringsSep "," ([
+            "vu"
+            "socket=${socket}"
+            "tag=${tag}"
+          ] ++ lib.optional dax
+            # dax_window is in bytes; omitting it means no DAX.
+            "dax_window=${toString (daxWindowSize * 1024 * 1024)}"
+          )))
         ] else throw "9p shares not implemented for alioth"
       ) shares
       ++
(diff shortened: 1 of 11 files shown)


lib/runners/alioth.nix
```diff
@@ -58,10 +58,14 @@ in {
...
+      builtins.concatMap ({ proto, socket, tag, dax, daxWindowSize, ... }:
         if proto == "virtiofs"
         then [
-          "--fs" (lib.escapeShellArg "vu,socket=${socket},tag=${tag}")
+          # dax_window is in bytes; omitting it means no DAX.
+          "--fs" (lib.escapeShellArg "vu,socket=${socket},tag=${tag}${
+            lib.optionalString dax
+              ",dax_window=${toString (daxWindowSize * 1024 * 1024)}"
+          }")
         ] else throw "9p shares not implemented for alioth"
```
SuperSandro2000, line 72: Please use a list concatenated with , here

SuperSandro2000 (commented):
otherwise LGTM

PR #601, opened 2026-09-11
qemu: do not enable test runner patch that forces p9 shares to be root owned

### LINKED ISSUE #599: all 9p shares appear root-owned in the guest VM (optimize.enable forces nixosTestRunner -> silently overwrites microvm.qemu.package)
**Symptom.** 
I used a plain 9p share and set `securityModel = "none"`. 
Still every file in the share is owned by `root:root` inside the virtual machine.
This behavior occurs regardless of host ownership, and files are read-only for unprivileged users in the vm.
Setting the option `microvm.qemu.package = pkgs.qemu_kvm` changes
nothing.

**Disclosure of AI usage**
I ran into this issue myself and did some research to resolve it using the LLM "Claude Fable 5".
I confirmed the fix myself and substantially reworked the LLM proposed issue report.

**Root cause.**
In `lib/runners/qemu.nix` the helper `minimizeQemuClosureSize` sets `nixosTestRunner` true if `oa.nixosTestRunner` is explicitly set true or graphics are disabled.
If `microvmConfig.optimize.enable` is true (the default), then `minimizeQemuClosureSize` is actually applied to the configured qemu package.

https://github.com/microvm-nix/microvm.nix/blob/804cbac7a462aa0fa8bb60c3d2fc4ead0a62060f/lib/runners/qemu.nix#L23-L35

In nixpkgs, the argument `nixosTestRunner` applies `force-uid0-on-9p.patch` which rewrites uid/gid to 0.
This is declared at

https://github.com/NixOS/nixpkgs/blob/f9532ae903bbbe53cab5914ab90e01953bcf1fce/pkgs/by-name/qe/qemu/package.nix#L288

https://github.com/NixOS/nixpkgs/blob/f9532ae903bbbe53cab5914ab90e01953bcf1fce/pkgs/by-name/qe/qemu/force-uid0-on-9p.patch#L52

It appears to me that this behavior exists mainly for the NixOS test driver.
There the host store is shared into test VMs and must appear as root-owned.
In different situations however, this behavior destroys ownership information.
I set `microvm.shares.*.securityModel = "none"`,
which the runner correctly serializes to security_model=none on the qemu command line (verified via ps);
the force-uid0-on-9p patch overrides it downstream, before any security-model branch.
This is in contrast to the behavior one expects after setting this option, as the documentation (https://www.qemu.org/docs/master/system/qemu-manpage.html) says:

> security_model=security_model
> 
>     Specifies the security model to be used for this export path. [...] In “passthrough” security model, files are stored using the same credentials as they are created on the guest. This requires QEMU to run as root. [...] “none” security model is same as passthrough except the sever won’t report failures if it fails to set file attributes like ownership.

**Reproduction** 
Use a config with a 9p share, graphics off and options to defaults otherwise:

```console
$ nix build .#nixosConfigurations.<vm>.config.microvm.declaredRunner \
    --no-link --print-out-paths
$ grep -o 'qemu[^/ ]*/bin/qemu-system' <runner>/bin/microvm-run
qemu-host-cpu-only-for-vm-tests-.../bin/qemu-system
```

Guest-side: `ls -ln` on the share shows `0 0` for every file;
create on the host a file and chown it to an arbitrary uid (e.g. 4321).
Observe that a guest-side `ls -ln` shows `0 0` as well.
Verified via strace on the qemu process that correct uids enter qemu
(`newfstatat` returns `st_uid=1000`) and are rewritten in the reply.

**Expected.** 
The options to `microvm.qemu` are honored as set in the configuration
and closure minimization does not change 9p semantics.
In particular file ownership is preserved when  `security_model=none` is set.

**Workaround.** 
Explicitly setting `microvm.optimize.enable = false;` restores correct ownership
(I confirmed this).

All claims tested at microvm.nix rev 174e28de151e0 (2026-08-30), nixpkgs nixos-26.05.
The mechanism is unchanged on current nixpkgs master (f9532ae, qemu 11.1.0) and current microvm.nix main.

SuperSandro2000:
Closes #599

will be cached with https://github.com/NixOS/nixpkgs/pull/562414

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1785692966,
-        "narHash": "sha256-vUfIeBEfpbAfZ5zjgIkYk7eHBeVfCYVjLbWnMkseYnk=",
+        "lastModified": 1789006805,
+        "narHash": "sha256-xB8mKMOx1IA9vTDNLmJZ6n4wCMq/cuWBBOzGCRnqxrU=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "643809054d65fdd466a63e3155b8c498cb483c04",
+        "rev": "8ce4ef6cb6f871616146b9fe26d2a5ae594e94fe",
         "type": "github"
       },
       "original": {
@@ -25,11 +25,11 @@
     "spectrum": {
       "flake": false,
       "locked": {
-        "lastModified": 1785761586,
-        "narHash": "sha256-MWOMVqJJERwjQGgRIv8d6rlEBqbLM3VZWxHkNKIIZNw=",
+        "lastModified": 1786032750,
+        "narHash": "sha256-+Ba/kIn5BjvV7PN8bWOo7RoqgTY/Yv2M7qeMf3wS2Ag=",
         "ref": "refs/heads/main",
-        "rev": "a7762d6f54b40560dd5255ce902e6e6a5d980fe9",
-        "revCount": 1416,
+        "rev": "77d2389874bb092b8dbf05ecc5b1030372e80e98",
+        "revCount": 1417,
         "type": "git",
         "url": "https://spectrum-os.org/git/spectrum"
       },
--- a/lib/runners/qemu.nix
+++ b/lib/runners/qemu.nix
@@ -20,13 +20,10 @@ let
     ];
   });
 
-  minimizeQemuClosureSize = pkg: pkg.override (oa: {
+  minimizeQemuClosureSize = pkg: pkg.override {
     # standin for disabling everything guilike by hand
-    nixosTestRunner =
-      if graphics.enable
-      then oa.nixosTestRunner or false
-      else true;
-  });
+    minimal = !graphics.enable;
+  };
 
   overrideQemu = x: lib.pipe x (
     lib.optional requireUsb enableLibusb


RooSoft:
This is affecting some web server microVM I am running... apparently they now run wayland, somehow...

justsomehumanbeing:
Just leaving this here from the #599 side so the option doesn't get lost.
This suggestion is untested beyond reading the source:
`nixosTestRunner` in qemu's `package.nix` (https://github.com/NixOS/nixpkgs/blob/f9532ae903bbbe53cab5914ab90e01953bcf1fce/pkgs/by-name/qe/qemu/package.nix) only gates ten GUI/audio flags (alsa, pulse, pipewire, jack, sdl, gtk, vnc, smartcard, spice, ncurses) plus the 9p patch (the `force-uid0-on-9p` patch mentioned in #599).
On the other hand `minimal` additionally drops `libslirp`, `seccompSupport`, the guest agent, blobs, docs and more. 

So disabling exactly those ten flags via `pkg.override { … = false; }` might give the old headless closure back without either regression.

If there's interest I can try to find time to test this in the next two weeks.
But since I am quite busy atm I cannot make promises, and if someone gets to it first, even better.

RooSoft:
This is my new configuration instead of minimal = true

gtkSupport = false;
sdlSupport = false;
spiceSupport = false;
vncSupport = false;
ncursesSupport = false;
openGLSupport = false;
alsaSupport = false;
pulseSupport = false;
pipewireSupport = false;
jackSupport = false;
smartcardSupport = false;

SuperSandro2000:
Maintaining those ten flags is to much maintainenace work for the added little benefit.

